PR #398, opened 2026-07-06
Interactions API revision 2026-05-20 migration + native Antigravity bridge

evansenter:
# Summary

Three major workstreams (12 commits, design docs included):

## 1. Interactions API — `Api-Revision: 2026-05-20` migration (breaking)

- Steps-based response model (replaces `Turn`), new SSE lifecycle (`step.start`/`step.delta`/`step.stop`), streaming function-call arguments, per-step usage, typed citations.
- Nine wire-format corrections derived from the generated SDK bindings and **verified against the live API**: lowercase enums, snake_case usage fields, etc.
- New surface: webhooks resource (CRUD + `:ping` + `:rotateSigningSecret`), per-request `webhook_config`, environments (remote sources + network allowlist), agents resource, retrieval tool (Vertex AI Search / RAG store / Exa / Parallel), typed `response_format` union (text/audio/image/video + inline|uri delivery), video generation config, multi-speaker TTS, deep-research options.

### Live verification results

Full integration suite run against the real API: **1308/1308 passing** (after fixes; two known-flaky tests hardened). Live testing caught four things fixtures could not:

| Finding | Fix |
|---|---|
| `function_call` steps carry a `signature` required on stateless replay (Google's generated bindings omit it) | `Step::FunctionCall`/`FunctionResult` gain `signature: Option<String>` |
| Response modalities are case-sensitive lowercase (`"IMAGE"` now rejected) | Builders send lowercase; `with_response_modalities` normalizes |
| `response_mime_type` rejected in **all** forms, even paired with `response_format` | Field removed outright (BREAKING); rejection still covered by a raw-JSON test |
| `webhook_config` requires `background=true` | Documented + test fixed |

Remaining fixture-only surface is tracked in `docs/INTERACTIONS_API_GAP.md`.

## 2. Native Antigravity bridge (`antigravity` feature)

A pure-Rust protocol client for the `localharness` agent runtime shipped in the `google-antigravity` wheel — no Python, no PyO3 (rationale in `docs/ANTIGRAVITY_BRIDGE_DESIGN.md`: the Go binary is the entire agent; the Python SDK is a thin protocol client, which is exactly the layer a Rust crate wants to own natively).

- Stdio protobuf handshake + proto-JSON WebSocket event pump, Evergreen unknown-event preservation, pinned `SUPPORTED_HARNESS_VERSION = "0.1.5"`.
- `#[tool]` functions work unchanged in both API and bridge modes; hooks and allow/deny policies are Rust closures; MCP server passthrough; structured output; session resume; streaming deltas.
- Client-side trigger scheduler (`add_trigger`): per-trigger timers injecting `automated_trigger` events, defer-until-idle with missed-interval collapse, abort-on-shutdown.
- Subagent registration (`add_subagent`): tool references validated against the parent registry at spawn; nested subagents force-disabled per harness semantics.

## 3. Wire debugging & modernization

- Structured `WireEvent`/`WireInspector` API with a `tracing` forwarder (`RUST_LOG=genai_rs::wire=debug`); `LOUD_WIRE=1` output preserved byte-for-byte.
- reqwest 0.13; two latent bugs fixed along the way.

# Test plan

- `make check` — 1027 unit tests
- `cargo nextest run --features antigravity` — 1126; `--features strict-unknown` — 947
- Doctests, `cargo doc -D warnings` (both feature sets), clippy `-D warnings` incl. `--features antigravity`, examples — all clean
- Live integration suite (real `GEMINI_API_KEY`): 1308/1308, including real agent chats through the harness
- Keyless harness suite (pinned wheel binary): 8/8 — now also run in CI via the new `test-antigravity` job

# Notes for review

- Breaking changes are deliberate per the crate's versioning philosophy (clean breaks over shims); CHANGELOG has migration notes. This is `0.8.0` material.
- Highest-value review targets: `src/antigravity/protocol.rs` + `handshake`/process spawn (new trust boundary), `src/steps.rs` signature handling, the policy/hook engine.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MqkKDDoTDKUp9zBk4M8WGH)_


evansenter:
All review findings addressed in f7c76dc. Per-finding disposition (every finding was independently verified against the source before acting):

**Important findings — all confirmed, all fixed:**

1. **Handshake allocation**: frames now capped at 4 MiB before allocation, with a clear "likely not a localharness" error. Unit-tested including the ~4 GiB prefix case.
2. **Turn-timeout desync**: the timeout path now sends `HaltRequest` and drains the abandoned turn's events (answering protocol-required confirmations so the harness can't deadlock) before returning `Timeout`. Keyless harness regression test asserts the wire order `userInput → haltRequest`.
3. **Deny fail-open**: verified against the pinned 0.1.5 proto descriptor that `ToolConfirmationRequest` has **no wire discriminator** (zero fields — the step's action fields are the only signal). Fix: steps with no action and empty Evergreen `extra` are genuine pre-request notifications and still auto-approve (the concrete call gets its own policy check); steps whose action landed in `extra` (unknown builtin) now **fail closed** unless the policy is allow-all, an exact rule names the wire field, or a pre-tool hook allows — with a `warn!` either way. This is deliberately stricter than the reference SDK; documented in `docs/ANTIGRAVITY.md`. Full decision-matrix unit tests added (finding 12).
4. **Trigger-turn desync**: fixed via a stronger form of the drain option — `chat()`/`send_streaming()` detect a fired trigger and **halt + drain** the trigger turn (covers in-flight, not just completed, turns) before sending user input. Trigger-turn output remains unsurfaced for now; documented in the module docs, `ANTIGRAVITY.md`, and the design doc records the background-consumer follow-up. Keyless regression test asserts `automatedTrigger → haltRequest → userInput`.
5. **reqwest 0.13 TLS roots**: confirmed the behavior change (Cargo.lock: `webpki-roots` ×3 → 0, `rustls-platform-verifier` added). However the suggested `rustls-tls-webpki-roots` feature **no longer exists in reqwest 0.13** — both rustls features hard-depend on `rustls-platform-verifier`, so parity can't be restored by a feature switch. The trust-anchor change is now documented in the CHANGELOG's reqwest entry. Restoring bundled Mozilla roots would mean adding a `webpki-roots` dependency and a preconfigured rustls `ClientConfig` — flagging that as a maintainer decision rather than doing it unilaterally.

**Suggestions:**

6–10, 12: all confirmed and fixed — media deltas now append to the last block like audio; `interaction.completed` usage falls back to the last `step.stop` cumulative usage; `RemoteEnvironment`/`EnvironmentSource`/`AllowlistEntry` preserve unknown fields via the repo-standard `#[serde(flatten)] extra`; `api_key` is redacted from LOUD_WIRE/tracing output (new `REDACT_FIELDS` — truncation was the wrong tool since keys are short); webhook signing secrets get manual `Debug` redaction per the `HttpContext` precedent; and the deny/allow confirmation decision matrix is unit-tested.

11. **`step_key` collision — refuted**: the harness emits proto3 JSON, which omits default-valued scalars, so an absent `step_index` *is* index 0 by protocol semantics; `unwrap_or_default()` is the correct decoding and two "None-index" steps in one trajectory are the same step. The assumption is now documented at the key-construction site.

Verification: `make check` 1039 passed; antigravity suite 1150; strict-unknown 959; keyless harness 10/10 (the two new regression tests re-run 3× stable); clippy 1.96 `--all-features -D warnings`, doctests, and docs all clean. The three key-gated harness chat tests also ran real turns exercising the new drain/confirmation logic.

---

evansenter:
Final round complete (`010b07f`…`eec07bd`). This PR is now **release-ready as 0.8.0**.

**Security review** (full-branch, adversarially verified): two MEDIUM findings, both secret-redaction gaps, both fixed — webhook signing secrets now in `REDACT_FIELDS` with `TracingForwarder` gaining redaction parity (it previously logged all bodies raw), and the antigravity WS inspection copy now masks `env`/`headers`/`httpHeaders` credential maps, not just `apiKey`. No HIGHs; the subprocess/protobuf/policy trust boundaries came back clean.

**Explicit live correctness verification** (LOUD_WIRE wire diffs + output sanity checks, not just pass/fail):
- *Core surface*: 10 flows byte-verified — after fixing two modeling gaps (`InteractionResponse.object`/`service_tier` silently dropped; `google_search_result` fabricating empty `title`/`url` on replay), every captured live response roundtrips byte-identically.
- *Phase-2 surface*: everything previously fixture-only is now live-verified. Notable server behaviors documented in code/docs: webhook `update_mask` is accepted but **ignored** (PATCH applies body fields exactly); `:ping` takes `{}`; rotation issues a fresh secret (old expires in 24h); agent `tools` limited to `code_execution`/`google_search`/`url_context`; image output locked to inline JPEG; audio locked to inline `audio/l16` + `sample_rate`; `include_input` is a live no-op; `Tool::Retrieval` + `enable_bigquery_tool` are Vertex-only; no Interactions model supports video today (task enum server-validated, revealing `extend`).

**New example project**: `examples/real_world/repo_auditor` — an antigravity-bridge agent auditing a deliberately-vulnerable fixture under `deny_all` + selective allows, with a subagent, a shared `#[tool]` classifier, and structured output cross-checked deterministically. Verified with real runs: 3/3 planted vulnerabilities found, zero false positives, `.env` read provably denied on the wire. Ergonomics friction from building it is captured in the design doc's new Follow-ups section.

**0.8.0 release prep**: version bump across the workspace, CHANGELOG consolidated with a 0.7→0.8 migration guide (steps model + signature replay, lowercase modalities, `response_mime_type` removal, speech_config list, reqwest 0.13 trust-store change), README rewritten for the current crate with doctested samples and an honest Vertex-gated-features section.

CI note: the core-group failure on `4a2bd8c` was a transient API latency spike (two stateful tests hit exact 60s/120s client timeouts; both pass locally in seconds) — this push re-runs the suite.

---

evansenter:
Second review round addressed in `dac165c`. All five findings were verified against the code and **all five confirmed** — good catches, particularly #1 and #4:

1. **Redaction recursion gap — fixed as suggested.** The `data`/`signature` truncation branch now recurses into non-string values, so secrets nested inside Evergreen `Unknown` payloads (which preserve raw JSON under `data`) can no longer escape `REDACT_FIELDS`. Regression tests cover both an object and an array under a `data` key.

2. **`Cancelled` arm `is_main` guard — fixed.** Chose the no-op guard (mirroring the `Idle` arm) rather than surfacing a new event type: subagent cancellations already reach the caller through step errors, so a dedicated `AgentEvent` variant would change the public enum for information already delivered. The routing was extracted into an associated fn to make it unit-testable — five new tests cover main/subagent × idle/cancelled.

3. **stderr drain — fixed as suggested.** `read_until(b'\n')` + lossy UTF-8 decode replaces `Lines`; invalid bytes no longer stop the drain (verified by a test that pushes garbage mid-stream and asserts later lines still flow). EOF and genuine I/O errors still end the loop; wire events, tracing forwarding, and the bounded ring are preserved.

4. **Trigger TOCTOU — confirmed and fixed with the mutex option.** A `turn_sync` mutex now makes the trigger task's [idle re-check → send] and `chat()`'s [`TurnGuard::begin` → `fired` swap] mutually exclusive; the trigger re-checks idleness under the lock and defers if a turn began meanwhile. The halt/drain deliberately stays outside the critical section so trigger tasks never block on a long drain (safe: the agent is busy, so triggers defer). Chose the mutex over a generation counter because it maps directly onto the two existing critical sections. Added a deterministic interleaving test (paused-time), and the three live-key harness chat tests were re-run green through the modified turn-begin path.

5. **Allowlist sibling keys — fixed** with the same flattened-`extra` pattern the file already uses for `RemoteEnvironment`; roundtrip test added. Not breaking: `NetworkConfig` is new in the unreleased 0.8.0.

Gates: `make check` 1048; antigravity 1168; strict-unknown 968; doctests 227; clippy 1.96 all-features and docs clean; harness suite 10/10 including live chat tests. CHANGELOG entries added inside the 0.8.0 section (finding 1 under Security).

---

PR #399, opened 2026-07-08
Antigravity bridge ergonomics: workspace announcement + faithful event stream

evansenter:
# Summary

Follow-up to the Antigravity bridge (#398), implementing the ergonomics deferred in `docs/ANTIGRAVITY_BRIDGE_DESIGN.md`'s **Follow-ups** list — the friction discovered while building the `repo_auditor` example. Five self-contained items; the larger deferred work (background consumer for trigger turns, mockable transport trait, async hooks) is left for a subsequent PR.

## What changed

1. **Workspace announcement.** `add_workspace(..)` roots are now announced to the model automatically — `spawn()` composes a concise, delimited note listing the absolute workspace root(s) onto the *effective* system instructions (the string passed to `with_system_instructions` is never mutated), and appends the same note to each subagent's instructions. Agents no longer guess paths like `/workdir`. Opt out with `with_workspace_announcement(false)`. **The wire protocol has no native announcement field** (verified against the harness descriptor), so this is prompt-level grounding.

2. **`ToolOutcome.result` unwrapped.** Post-tool hooks previously received the raw `{"result": …}` wire envelope; they now see the tool's inner return value.

3. **Accepted/denied marker on tool actions.** `AgentEvent::ToolAction` is now a struct variant carrying a `ToolDecision` (`Allowed` / `Denied { reason }`), so a policy- or hook-denied action no longer looks identical to an executed one in the event stream.

4. **Trajectory identity + subagent name.** The `ToolAction` event carries a `trajectory_id` for source attribution, and `ToolAction::subagent_name()` surfaces the delegated subagent's name from `InvokeSubagent` (previously only reachable via Evergreen `extra`).

5. **Error severity classification.** `AgentEvent::Error` is now a struct variant with an `ErrorSeverity`, so transient harness noise (retried internally, turn continues) is distinguishable from turn-ending errors.

Items 3–5 reshape a few public `antigravity` enums — clean breaks, acceptable since 0.8.0 is merged but not yet published to crates.io. The `repo_auditor` example is updated to use the new surface and drops its manual workspace-in-prompt workaround.

## Verification

- `make check` 1048; `--features antigravity` 1181; `--features strict-unknown` 968; doctests, `cargo doc -D warnings` (all-features), clippy `-D warnings` (all-features), and `cargo check --examples` all clean.
- Keyless + key-gated harness suite: 10/10.
- **`repo_auditor` run live end-to-end**: finds all 3 planted vulnerabilities with the deterministic severity cross-check passing. The LOUD_WIRE trace confirms the workspace root is announced in the init system instructions (item 1) and the agent uses exact absolute paths with no guessing; the denied `.env` read surfaces as `[DENIED view_file]` (item 3); and the post-tool hook prints the unwrapped `{"category":…,"severity":…}` value (item 2).

## Still open (future follow-up)

Background consumer for idle trigger turns, a mockable transport trait for unit-testing the turn loop, and async hooks — all noted in the design doc's Follow-ups section.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MqkKDDoTDKUp9zBk4M8WGH)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `ActionInvokeSubagent::name` field (Evergreen; `None` on harness 0.1.5,
   which emits an empty `invokeSubagent` action).
 - **Antigravity: `ToolDecision`** (`Allowed` / `Denied { reason }`) and
-  **`ErrorSeverity`** (`Transient` / `Terminal`) public enums.
+  **`ErrorSeverity`** (`Transient` / `Severe`) public enums.
 
 ### Changed
 
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Antigravity (breaking): `AgentEvent::Error(String)`** is now
   `Error { message, severity }` so consumers can ignore transient
   harness-internal noise (retried internally; the turn continues) and react
-  only to `Terminal` errors. Turn-ending failures still surface as
+  only to `Severe` errors. Turn-ending failures still surface as
   `AntigravityError::Turn`.
 - **Antigravity (behavior): `ToolOutcome.result`** passed to post-tool hooks
   is now the inner tool result, not the raw `{"result": ...}` wire envelope
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Antigravity: workspace announcement.** `add_workspace(..)` roots are now
+  announced to the model automatically — `spawn()` prepends a concise,
+  delimited note listing the absolute workspace root(s) to the effective
+  system instructions (the string passed to `with_system_instructions` is
+  never mutated), and appends the same note to every subagent's instructions.
+  Agents no longer guess workspace paths. Opt out with
+  `AgentBuilder::with_workspace_announcement(false)`. The wire protocol has no
+  native announcement field, so this is prompt-level grounding.
+- **Antigravity: `ToolAction::subagent_name()`** and a typed
+  `ActionInvokeSubagent::name` field (Evergreen; `None` on harness 0.1.5,
+  which emits an empty `invokeSubagent` action).
+- **Antigravity: `ToolDecision`** (`Allowed` / `Denied { reason }`) and
+  **`ErrorSeverity`** (`Transient` / `Terminal`) public enums.
+
+### Changed
+
+- **Antigravity (breaking): `AgentEvent::ToolAction`** is now a struct variant
+  `ToolAction { action, decision, trajectory_id }`. `decision` distinguishes
+  executed actions from policy/hook-denied ones (previously indistinguishable
+  in the stream), and `trajectory_id` tells parent and subagent actions apart.
+- **Antigravity (breaking): `AgentEvent::Error(String)`** is now
+  `Error { message, severity }` so consumers can ignore transient
+  harness-internal noise (retried internally; the turn continues) and react
+  only to `Terminal` errors. Turn-ending failures still surface as
+  `AntigravityError::Turn`.
+- **Antigravity (behavior): `ToolOutcome.result`** passed to post-tool hooks
+  is now the inner tool result, not the raw `{"result": ...}` wire envelope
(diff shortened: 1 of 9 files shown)


evansenter:
Thanks for the review. Both suggestions addressed:

**1. "absolute path(s)" wording (`19c5392`):** softened to "the following path(s)". I went with the wording option rather than canonicalizing on purpose — `fs::canonicalize` resolves symlinks, so the announced path could diverge from what the harness echoes in `list_directory`/`view_file` results, which would confuse the model *more* than a relative path. Announcing exactly the string the harness roots its tools at keeps them consistent. Updated the builder doc comment to match.

**2. denied-action dedup test coverage:** agreed, and left as-is — as you noted, the wiring is exercised by the live `repo_auditor` run but not by a unit test, because there's no mockable transport yet. That transport trait is already on the deferred Follow-ups list; I'll add the direct `answer_tool_confirmation`/`handle_step` dedup-keying test in the same PR that lands it, so the two arrive together.

---

evansenter:
Good catch — fixed. The note is emitted as the second system-instruction part (after the user's text), so "prepends" was wrong; corrected to "appends" in the builder doc, `docs/ANTIGRAVITY.md`, `CHANGELOG.md`, and the design doc (all four occurrences). Also dropped the stale "absolute" wording that lingered in the two docs, to match the announcement softening from the earlier round. Doc-only change; behavior unchanged.

---

evansenter:
Both non-blocking suggestions from the last two reviews addressed in `4388085`:

**Error-step dedup:** you're right that the reshaped `AgentEvent::Error` block had no dedup while the adjacent tool-action path does. Since an error step is terminal and the harness re-delivers it each tick, that could surface duplicate `Error` events and inflate `turn.errors`. Added a guard — but using a **dedicated `announced_errors` set** rather than reusing `announced_actions`: a single step can carry both a tool action and an error (e.g. a `run_command` that failed), and sharing the set would let the action's announcement suppress the error. Now each surfaces exactly once. The fatal-abort path is untouched (it returns before the dedup).

**Doc sketch:** updated the `send_streaming` example in `docs/ANTIGRAVITY_BRIDGE_DESIGN.md` to the new `AgentEvent::ToolAction { action, decision, .. }` struct-variant form.

Also, re the `Integration Tests (core)` failure webhook: that was `test_conversation_branch`, a live semantic-validation multi-turn test on the Interactions HTTP API — a path this PR doesn't touch. It passes locally (26s); it's the usual double-LLM flake, and this push re-runs it.

Verification: antigravity suite 1181; clippy `--all-features`, fmt, docs clean.

---

evansenter:
Both suggestions from this round addressed in `9f5ac58`:

**Denial/dedup unit test.** You're right that the send-side I/O isn't actually in the way of locking this. I extracted the once-per-step dedup into `TurnState::announce_tool_action`, now called by *both* the terminal-`Allowed` site and the confirmation-`Denied` site (so the two can't drift), and added `test_denied_action_dedups_terminal_allowed_emission`: it announces `Denied` for a step_key, asserts a later `Allowed` for the same key is suppressed, that exactly one (Denied) event is queued, and that a distinct step_key is unaffected. No mock transport needed — the key is computed once in `process_step` and handed to `answer_tool_confirmation`, so both sites share it structurally. The broader `process_step`-driven test (with a real/mock session) still belongs with the mock-transport follow-up, but the dedup invariant itself is now locked.

**`ErrorSeverity::Terminal` → `Severe`.** Renamed. I went with `Severe` over your other suggestion `Fatal`, because "fatal" carries the same "it's over" connotation that made `Terminal` misleading — and this variant explicitly does *not* end the turn. Updated all usages, the docs, the CHANGELOG, and a now-inaccurate comment in the `repo_auditor` example that had claimed these errors abort the turn.

Verification: antigravity suite 1182; clippy `--all-features`, fmt, docs clean.

---

evansenter:
Fixed in `552ffc9` — you're right, my rename sweep's grep didn't cover `ANTIGRAVITY_BRIDGE_DESIGN.md`, so line 336 kept the old `(Transient / Terminal)`. Now `(Transient / Severe)`. Confirmed the only remaining `Terminal` string in the antigravity surface is the deliberate note in `streaming.rs` explaining why the variant *isn't* named that.

Separately: the `Integration Tests (tools)` red on `9f5ac58` was `test_google_search_streaming`, a live server-side Google Search grounding test unrelated to this branch's diff — it hit a streaming hiccup and passes locally (17.8s). The API job-rerun is blocked for this token (403), so `32bd596` is an empty commit to re-trigger the matrix; this doc fix rides on top.

---

PR #416, opened 2026-08-15
Release 0.10.0

evansenter:
Version bump for the release that lands #415, plus a CI fix this PR ran into.

## Why minor, not patch

Cargo treats the minor as the major for pre-1.0 crates. The README's
installation section tells people to write `genai-rs = "0.9"`, which matches
any 0.9.x — so a 0.9.1 would reach every existing user on their next
`cargo update` with no opt-in. 0.10.0 makes them edit the dependency line,
which is the friction a behavior break should carry.

## The break

`AgentBuilder`'s per-turn budget now defaults to `DEFAULT_TURN_TIMEOUT`
(300s) instead of unlimited. It compiles fine and fails at runtime: an agent
whose turn legitimately runs past 300s — a deep subagent tree, a long tool
chain — now gets `AntigravityError::Timeout` where it previously completed.
`without_turn_timeout()` restores the old behavior, which is itself the
admission that this is a behavior choice rather than a defect being fixed.

Everything else in the release is additive (the five model constants),
internal (the model-literal guard, the CI vacuous-pass guards), or cosmetic
(the examples move, docs).

## Changes

| File | Change |
|------|--------|
| `Cargo.toml` | `version`, and the `genai-rs-macros` dependency constraint |
| `genai-rs-macros/Cargo.toml` | `version` |
| `CHANGELOG.md` | `[Unreleased]` → `[0.10.0] - 2026-08-16` |
| `Cargo.lock` | both workspace crates, as cargo regenerated it |
| `README.md` | both installation lines → `"0.10"` |
| `docs/ANTIGRAVITY.md` | setup snippet → `"0.10"` |

The `rand = "0.9"` line in `Cargo.toml` is an unrelated dependency and is
deliberately untouched.

## CI: bound the mold install

`sudo apt-get install -y mold` runs in eleven steps across `rust.yml` and two
more in `release.yml`, and none was bounded. A stalled apt mirror does not
fail the command, it hangs — and with no ceiling the runner is held until
GitHub's 6h default.

This PR hit exactly that: the Test Suite job sat **98 minutes** on that step
and never reached the tests, while every other job went green, so the PR
showed one pending check with nothing to read. A cancel and re-run cleared it
in 9 seconds.

All thirteen steps now carry `timeout-minutes: 5` — normally 10–20s, so the
ceiling only trips on a stall, and the failure names the step instead of
stalling silently. Step-level rather than job-level, matching the existing
timeout on the antigravity example smoke-run: it points at the step that
actually hangs and leaves the long jobs their full budget.

`release.yml` also used `apt-get update && install`, so a transient mirror
error would fail the release outright where `rust.yml` proceeds; it now uses
`|| true;` to match. That relaxation is scoped to `update` alone — the
install remains the last command under `bash -e`, so a genuinely missing
package still fails loudly.

No CHANGELOG entry for the CI work — CI-only, per the repo's convention.

## Verification

Run against the bump tree locally: `cargo fmt --check` clean, `cargo clippy
--workspace --all-targets --all-features -- -D warnings` clean, **1198/1198
tests passing**, `RUSTDOCFLAGS="-D warnings" cargo doc` clean. In CI, the
full suite is green on `6a0f03d` including all five integration groups and
both doctest steps; the workflow change is validated by CI running it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,13 @@ jobs:
       - uses: taiki-e/install-action@cargo-nextest
       - uses: Swatinem/rust-cache@v2
       - name: Install mold linker
-        run: sudo apt-get update && sudo apt-get install -y mold
+        # Same stall ceiling as rust.yml's mold installs, and it matters more
+        # here: a stall holds publishing after the tag already exists. The
+        # `|| true` also matches rust.yml — a transient mirror error on
+        # `update` should not fail the release when the install can still
+        # proceed from cached lists.
+        timeout-minutes: 5
+        run: sudo apt-get update || true; sudo apt-get install -y mold
       - name: Check workspace
         run: cargo check --workspace --all-targets --all-features
       - name: Run unit tests
@@ -101,7 +107,13 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - name: Install mold linker
-        run: sudo apt-get update && sudo apt-get install -y mold
+        # Same stall ceiling as rust.yml's mold installs, and it matters more
+        # here: a stall holds publishing after the tag already exists. The
+        # `|| true` also matches rust.yml — a transient mirror error on
+        # `update` should not fail the release when the install can still
+        # proceed from cached lists.
+        timeout-minutes: 5
+        run: sudo apt-get update || true; sudo apt-get install -y mold
 
       - name: Verify tag matches Cargo.toml version
         run: |
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,6 +19,12 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - name: Install mold linker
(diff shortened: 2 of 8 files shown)


evansenter:
Both review findings were already fixed when the review posted — it ran against `1548be3`, and `75b7f80` landed first. Its own summary ("the four edited files") describes that earlier four-file state.

On the current head `75b7f80`:

```
Cargo.lock:432   version = "0.10.0"    # [[package]] name = "genai-rs"
Cargo.lock:465   version = "0.10.0"    # [[package]] name = "genai-rs-macros"
CHANGELOG.md:8   ## [0.10.0] - 2026-08-15
```

The lockfile was regenerated by cargo, not hand-edited. The two remaining `0.9.0` entries in `Cargo.lock` are `rand_chacha` and `untrusted` — unrelated third-party dependencies.

All six rows of the CLAUDE.md Version Bump Checklist are now covered, so the `--locked` inconsistency the verdict is based on does not apply to this head. The blocking review can be dismissed.

The split happened because the first four files went in through the GitHub API while this session's git push credentials were briefly unavailable; the last two followed by git once they recovered.

---

Issue #429, opened 2026-08-16
`has_file_search_results()` is true while `file_search_results()` is empty — the API returns no chunks


evansenter:
Found while implementing #422. The two accessors disagree in a way that will mislead callers, though neither is wrong on its own terms.

## What the API returns

A `file_search_result` step contains only three keys — there is no `result` field:

```
KEYS: ['call_id', 'signature', 'type']
result field: null
```

Captured live 2026-08-16 (`gemini-3.7-flash`, Api-Revision `2026-05-20`) from an interaction that *did* retrieve successfully — the model answered correctly from the uploaded document, so retrieval happened; the chunks just aren't surfaced in the response.

## The resulting mismatch

```rust
// src/response.rs
pub fn has_file_search_results(&self) -> bool {
    self.steps.iter().any(|s| matches!(s, Step::FileSearchResult { .. }))
}

pub fn file_search_results(&self) -> Vec<&FileSearchResultItem> {
    // flattens each step's `result` vec — all empty
}
```

`has_*` asks "is there a result step?" and `file_search_results` asks "what chunks are in them?". Both answer their own question correctly, but a caller reasonably reads the pair as "if has, then non-empty" and gets nothing:

```
Steps: ["file_search_call", "file_search_result", "thought", "model_output"]
Retrieved chunks exposed by the API: 0
```

`FileSearchResultItem` (title / store / text) is presumably modeled from the spec rather than from an observed payload.

## Options

1. **Document it and leave the semantics alone** (what #422 shipped — a note on the example and in `ENUM_WIRE_FORMATS.md`). Honest, zero risk, but the trap stays for anyone reading only rustdoc.
2. **Add doc comments on both methods** stating that the API currently returns no chunk contents, so `has_*` means "a search occurred", not "chunks are available". Cheapest real fix; probably worth doing regardless of what else is chosen.
3. **Redefine `has_file_search_results()` as `!file_search_results().is_empty()`.** Makes the pair consistent, but then nothing reports that a search occurred — which is the more useful signal today, and what the integration test asserts on.
4. **Split the two questions**: keep `has_file_search_results()` as-is and add `file_search_call_count()` or similar, so "a search ran" and "chunks came back" are separately askable.

Leaning 2 now, and 4 if the API starts populating `result`. Option 3 loses information.

## Worth watching

If the API begins returning chunks, `file_search_results()` starts working with no code change and the doc note becomes wrong. Whatever lands here should be re-checked as part of the recurring sweep in #421.

Issue #307, opened 2026-01-08
Add tests for File Search tool when stores are available


evansenter:
## Summary

The `with_file_search()` and `with_file_search_config()` builder methods exist but have no integration tests.

## Current State

- `with_file_search()` - implemented in `src/request_builder/mod.rs:1594`
- `with_file_search_config()` - implemented in `src/request_builder/mod.rs:1638`
- Wire format documented in `docs/ENUM_WIRE_FORMATS.md`
- No tests exist

## Why Blocked

File Search requires configured file search stores, which requires:
1. Creating file search stores via the API
2. Uploading files to those stores
3. Then querying them

This is different from the Files API - file search stores are a separate concept.

## When to Unblock

- When we have a test harness for creating/managing file search stores
- Or when Google provides a simpler way to test file search

## Action Items

- [ ] Research file search store creation process
- [ ] Add unit tests for builder methods
- [ ] Add integration test with test store (if feasible)
- [ ] Document file search setup process

evansenter:
**Unblocked** — File Search stores are available on a standard `GEMINI_API_KEY`. Full lifecycle verified 2026-08-16:

| Call | Result |
|---|---|
| `POST /v1beta/fileSearchStores` | 200 — returns `name`, `displayName`, `createTime`, `embeddingModel: "models/gemini-embedding-001"` |
| `GET /v1beta/fileSearchStores` | 200 |
| `GET /v1beta/fileSearchStores/{id}/documents` | 200 |
| `POST /upload/v1beta/{store}:uploadToFileSearchStore` (multipart) | 200 — `documentName`, `mimeType`, `sizeBytes` |
| `DELETE /v1beta/fileSearchStores/{id}?force=true` | 200 |

End-to-end retrieval confirmed: uploaded a document containing a unique token to a fresh store, queried it through `Tool::FileSearch` on `gemini-3.7-flash`, and got `status: completed` with steps `file_search_call` → `file_search_result` → `thought` → `model_output` and the correct grounded answer. Probe store deleted afterward.

Removing the `blocked` label.

One dependency worth noting: the crate has no store-management surface, so a test would currently have to provision its store out-of-band. #422 covers adding `Client::*_file_search_store*()` — landing that first makes these tests self-contained rather than dependent on a manually created store.

Found during a stale-issue / feature audit.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #435, opened 2026-08-16
Add File Search Store management (/v1beta/fileSearchStores)

### LINKED ISSUE #422: No File Search Store management surface — the `file_search` tool is unusable from the crate alone
`Tool::FileSearch` takes `store_names`, but `Client` has no way to create a store or put documents into one. A caller using only this crate cannot reach a working file-search setup — they have to shell out to `curl` or the Python SDK to provision the store first, then hand the name back to us.

The store API is live today. Full lifecycle verified 2026-08-16 with a standard `GEMINI_API_KEY`:

| Call | Result |
|---|---|
| `POST /v1beta/fileSearchStores` `{"display_name": ...}` | 200 — returns `name`, `displayName`, `createTime`, `updateTime`, `embeddingModel: "models/gemini-embedding-001"` |
| `GET /v1beta/fileSearchStores` | 200 |
| `GET /v1beta/fileSearchStores/{id}/documents` | 200 |
| `POST /upload/v1beta/{store}:uploadToFileSearchStore` (multipart) | 200 — returns an operation with `documentName`, `mimeType`, `sizeBytes` |
| `DELETE /v1beta/fileSearchStores/{id}?force=true` | 200 |

End-to-end retrieval confirmed against a real interaction: uploaded a text document to a fresh store, queried it via `Tool::FileSearch`, and the response came back `status: completed` with steps `file_search_call` → `file_search_result` → `thought` → `model_output` and the correct answer grounded in the uploaded content.

## Proposed work

- [ ] `Client::create_file_search_store()` / `get_` / `list_` / `delete_` (with the `force` query param on delete)
- [ ] Document upload: multipart `:uploadToFileSearchStore`, mirroring the existing Files API upload methods in `src/client.rs`
- [ ] Document listing / deletion under a store
- [ ] Typed response shapes with a flattened Evergreen `extra` (note the store API returns **camelCase** — `displayName`, `createTime`, `embeddingModel` — unlike the Interactions API's snake_case; verify with `LOUD_WIRE=1` before settling the serde attributes)
- [ ] `with_file_search()` builder convenience (see the builder-consistency issue)
- [ ] Integration tests — this is what unblocks #307
- [ ] Extend `examples/file_search.rs` to provision its own store rather than assuming one exists

Note the upload response is an operation wrapper; check whether it can return unresolved and needs polling, similar to `wait_for_file_ready()`.

evansenter:
Closes #422. Unblocks #307.

`Tool::FileSearch` takes store names, but the crate had no way to *create* a store — so file search was unusable without provisioning one out-of-band. That is why #307 sat blocked.

Adds store CRUD, document upload/list/get/delete, and `wait_for_document_active`, plus `FileSearchStore`, `FileSearchDocument`, and `DocumentState` types with Evergreen `extra` passthrough.

## Wire details, all verified live 2026-08-16

- **This resource is camelCase** (`displayName`, `createTime`, `embeddingModel`), unlike the Interactions API's snake_case. The types carry an explicit rename.
- `sizeBytes` is a JSON **string**, routed through the existing protobuf-JSON int64 helpers.
- `DocumentState` uses a `STATE_` prefix (`STATE_PENDING`/`STATE_ACTIVE`), which the Files API's `FileState` does not.
- **Indexing is asynchronous.** A fresh upload is Pending and file search matches *nothing* until Active — it fails silently, returning a confident answer without the document. Hence `wait_for_document_active`.
- Deleting an indexed document, or a store holding documents, requires `force=true`.
- The upload endpoint accepts both `raw` and `multipart`. This uses `raw` (bytes as body, `display_name` as query param) so the crate need not enable reqwest's `multipart` feature for one call site.
- The upload response is an **operation wrapper**, not a document; `response.documentName` is resolved via a follow-up GET so callers get the same shape list/get return.

Resource names are validated locally with the same positive-shape check the Files API uses, so a malformed name fails as `InvalidInput` rather than misfiring at a wrong URL.

## Two findings that changed other things

**`file_search_result` steps carry no chunks** — only `call_id`, `signature`, `type`. So `has_file_search_results()` is `true` while `file_search_results()` is empty. Filed as #429.

**`file_search` and `google_search` cannot be combined** (`400 ... Please choose one to continue`). The previous example advertised that pairing as a "RAG + web grounding pattern".

## Example rewritten

`examples/file_search.rs` now provisions its own store and runs end to end, replacing a placeholder store ID that could never have worked. Three of its production notes were wrong and are corrected: store IDs are `fileSearchStores/<id>` not `stores/<id>`; the google_search pairing is rejected; results do not "include document title, text snippet, and source store".

## Testing

Six integration tests — store lifecycle, upload and indexing, force semantics, local name validation, and the end-to-end retrieval #307 wanted. **All green live:**

```
steps: ["file_search_call", "file_search_result", "thought", "model_output"]
retrieval answer: The maintenance codename for the Vega ground station relay is HALCYON-девять-42.
```

---


evansenter:
## Feedback Addressed

All six round-1 suggestions implemented. Three were more than polish:

**Tests unreachable from CI** — correct, and it undercut the PR's own premise: live verification is the evidence here, and only the two local-validation tests were running. Added `file_search_stores_tests` to `files-and-wire`.

**Cleanup only on the happy path** — the module doc claimed "every test cleans up the stores it creates" and that was false the moment any assertion failed. Bodies now run through a `with_store` helper that `catch_unwind`s, deletes, then resumes the panic — so a failing test still fails, and the claim is now true. `test_store_lifecycle` deliberately stays outside it, since deleting the store *is* its final assertion.

**`rejects_non_empty` didn't assert** — you were right that it passed identically either way while the docs stated the behavior as live-verified. It asserts now, and holds live: `Cannot delete non-empty FileSearchStore`.

Also applied: the Files API's empty/oversize guards on the upload path, the unresolved-operation error now recording what was actually observed (resolved on every upload through ~1 MB) and surfacing the operation handle, and the example logging its cleanup failure instead of `?`-ing over the original error.

All six integration tests re-verified live after the changes.

---

## Round 2

**Docs still teaching the corrected claims** — this is the one I most want to fix and it's the largest, so flagging intent rather than silently deferring: `BUILT_IN_TOOLS.md` has the `stores/my-store-123` prefix in four places, tells readers to provision via AI Studio, and iterates `file_search_results()` printing fields #429 records as never populated; `EXAMPLES_INDEX.md` still lists a pre-configured store as a prerequisite. All three are exactly what this PR disproved. Doing it in the next push.

**`get_document` naming** — agreed, and worth taking while these are unreleased. `get_document` reads as crate-level next to `get_file`/`get_agent`/`get_environment`.

**`poll_interval` parameter** — agreed, matches `wait_for_file_ready` and costs nothing since `timeout` is already `Option`al.

**Untested `_with_request` / `_with_mime`** — fair, and the `extra` passthrough is the part most likely to surprise since it rides `#[serde(flatten)]`. Will add the unit test and route one integration test through `_with_mime`.

---

Issue #425, opened 2026-08-16
CI flakiness report opens a new issue every day and never closes the old one (15 open, oldest 4.5 months)


evansenter:
There are currently **15 open `ci-health` flakiness reports** — #391-#397, #407-#408, #410-#414, #417 — spanning 2026-03-30 to 2026-08-16. The oldest have been open 4.5 months. This is a workflow bug, not a triage backlog.

## Root cause

`.github/workflows/ci-flakiness-report.yml` runs **daily**, but titles each report by day while calling it a week:

```
TITLE="CI Flakiness Report - Week of $(date -u +"%Y-%m-%d")"
```

and the dedupe check only matches an existing issue whose title starts with *today's* date:

```
EXISTING_ISSUE=$(gh issue list \
  --json number,title \
  --jq ".[] | select(.title | startswith(\"CI Flakiness Report - Week of $(date -u +"%Y-%m-%d")\")) | .number" ...)
```

So the update path fires only if the job runs twice in one day. On every ordinary day the match is empty and it creates a fresh issue. Nothing ever closes the previous one, so the count grows by one per day indefinitely.

The March/April cluster stopping at #397 and resuming at #407 in August suggests the scheduled job was disabled or failing for ~4 months and has since resumed — worth confirming, since it also means those two clusters describe long-stale CI state.

## Options

1. **One rolling issue** — fixed title (no date), always updated in place. Simplest; loses history.
2. **Genuinely weekly** — key the title on the ISO week (`date -u +%G-W%V`) and switch the schedule to weekly, matching what the title already claims.
3. **Daily but self-closing** — keep per-day issues, and close the prior one when opening the next.

Recommend option 1 or 2. Option 3 preserves the churn.

## Proposed work

- [ ] Fix the title/dedupe mismatch in `ci-flakiness-report.yml`
- [ ] Bulk-close the stale reports as `not_planned` — the March/April ones describe CI state that no longer exists
- [ ] Confirm whether the Apr-Aug gap was a disabled schedule or a silently failing job; if the latter, the job needs to fail loudly
- [ ] Consider whether an unactioned flakiness report should auto-close after N days rather than accumulate

evansenter:
Fixed on branch `claude/issue-425-flakiness-workflow`, uncommitted to remote pending review.

[...]

Issue #420, opened 2026-08-16
`generation_config.speech_config` widened to a union — crate models only the list arm


evansenter:
`google-genai` 2.18.x widens `generation_config.speech_config` from a plain list to a union. The crate models only the list arm, so the new object form fails to deserialize.

## Change between 2.17.0 and 2.18.1

`google/genai/_gaos/types/interactions/generationconfig.py`:

```diff
- speech_config: Optional[List[SpeechConfig]] = None
+ speech_config: Optional[SpeechConfigUnion] = None

+ SpeechConfigUnion = TypeAliasType(
+     "SpeechConfigUnion", Union[SpeakerConfig, List[SpeechConfig]]
+ )
+ """Optional. Speech and multi-speaker configuration."""
```

The new arm is `SpeakerConfig` (`speakerconfig.py`), an object wrapping the same per-speaker configs:

```
SpeakerConfig { speakers: Optional[List[SpeechConfig]] }
```

Current crate state: `pub speech_config: Option<Vec<SpeechConfig>>` in `src/request.rs`, with `with_speech_configs()` / `add_speech_config()`.

## Impact

Sending is fine — the list arm is still valid, so nothing breaks on the request side today. The risk is on **deserialize**: if the API starts echoing back (or a stored interaction returns) the `{"speakers": [...]}` object form, the crate errors instead of degrading. That is the failure mode the Evergreen philosophy exists to prevent.

## Proposed work

- [ ] Model the union as an enum with a `List` arm and a `Speaker { speakers }` arm, plus an `Unknown` variant with the standard three helpers
- [ ] Keep `with_speech_configs()` / `add_speech_config()` emitting the list form (no caller-visible break)
- [ ] Add a builder path for the object form if it proves to carry fields the list form cannot
- [ ] Live-probe both arms with `LOUD_WIRE=1` before choosing which one to emit by default — the object arm has not yet been verified against the Gemini API, only against the generated bindings
- [ ] Roundtrip tests for both arms; update `docs/ENUM_WIRE_FORMATS.md`

Found during a stale-issue / feature audit.

evansenter:
Implemented on branch `claude/issue-420-speech-config-union`, uncommitted to remote pending review. Two corrections to this issue's premises, both from the live probe its own checklist called for.

### 1. The failure mode is silent data loss, not a deserialize error

[...]

Issue #419, opened 2026-08-16
Video content `processing` field is unmodeled (static clipping / fps / agentic) — 127x token-cost difference


evansenter:
`google-genai` 2.18.x adds a `processing` field to video content that the crate does not model. `Content::Video` (`src/content.rs:1073`) carries only `data`/`uri`/`mime_type`/`resolution`, and the variant has no flattened `extra`, so the field is dropped in both directions.

## Wire shape

`processing` is a union:

- **Object form** — `{"type": "static", "start_offset": "10.5s", "end_offset": "30s", "fps": 1}`
  - `start_offset` / `end_offset`: decimal seconds with an `s` suffix; non-negative, end > start
  - `fps`: video frame-rate sampling density
- **Enum form** — `"static"` | `"agentic"` (open enum; the generated bindings use `UnrecognizedStr`)

Source: `google/genai/_gaos/types/interactions/videocontent.py`, `mediaprocessing.py`, `staticmediaprocessing.py` in 2.18.1. Absent in 2.17.0, which is where `docs/INTERACTIONS_API_GAP.md` last swept.

## Live verification (2026-08-16, `gemini-3.7-flash`, Api-Revision 2026-05-20)

Same YouTube URI in all three, varying only `processing`:

| `processing` | Result |
|---|---|
| `{"type":"static","start_offset":"5s","end_offset":"10s","fps":1}` | 200 — **455** video input tokens |
| `"agentic"` | 200 — **57,775** video input tokens |
| `"bogus_nonsense"` | 400 `Invalid enum value 'bogus_nonsense' at 'input[0].content[1].processing'` |

The rejection names the exact field path, confirming server-side validation rather than silent passthrough. The 127x spread between a clipped static window and agentic processing makes this a cost-control knob, not a cosmetic one — today a caller has no way to reach it.

## Proposed work

- [ ] Add `processing: Option<VideoProcessing>` to `Content::Video` (serialize + deserialize, snake_case)
- [ ] Model the union: object arm `{type: "static", start_offset, end_offset, fps}` and bare-string arm, with an `Unknown` variant per the Evergreen pattern (`is_unknown()` / `unknown_processing_type()` / `unknown_data()`)
- [ ] Builder support on the video content constructors
- [ ] Serialization roundtrip tests + a proptest strategy
- [ ] Update `docs/ENUM_WIRE_FORMATS.md` with the verified wire format above
[...]

PR #438, opened 2026-08-16
Mark the gap tracker as a snapshot and add a recurring API surface sweep

### LINKED ISSUE #421: `INTERACTIONS_API_GAP.md` asserts completeness it cannot back — needs a staleness marker and a recurring sweep
Two problems, and the second is the one that bites.

## 1. The tracker has drifted

`docs/INTERACTIONS_API_GAP.md` was swept against `google-genai` 2.17.0. 2.18.0 and 2.18.1 have since shipped, and the diff contains real unmodeled surface:

- `Content::Video.processing` — static clipping / fps / agentic modes (#419), live-verified with a 127x token-cost spread
- `generation_config.speech_config` widened to a `SpeakerConfig | List[SpeechConfig]` union (#420)

## 2. The tracker reads as authoritative when it is only self-reported

Every item in "Missing surface (by user value)" is struck through and the header says:

```
Status: implementation tracker — items are removed/checked off as they land.
```

Nothing in the file distinguishes "verified complete as of a sweep" from "complete". There is no upper bound on the claim and no marker for how stale it is. A reader — human or agent — who opens this file reasonably concludes the Interactions API surface is fully covered, which was false at the time of writing this issue.

This was demonstrated during the audit that produced #419/#420: the tracker was consulted, taken at face value, and the conclusion "surface fully covered" was drawn from it. That conclusion was wrong. Cross-checking against Google's own `ai.google.dev` reference did **not** catch it either — the published reference does not document `processing` at all. Only diffing the generated Python bindings surfaced it, and only a live probe confirmed it.

The ordering that actually worked, and that the doc should state:

1. **Generated bindings** (`google-genai` `_gaos/types/interactions/*.py`) — leading indicator; machine-generated from the spec, ships ahead of prose
2. **Live probes** — ground truth; the API validated `processing` and rejected a bogus enum value with an exact field path
3. **Prose docs** — both Google's reference and this tracker lag the other two

## Proposed work

- [ ] Add a prominent header: the SDK version last swept, the sweep date, and an explicit "this is a point-in-time snapshot, not a completeness guarantee"
- [ ] Document the verification ordering above in the file itself, so the next reader does not repeat the mistake
- [ ] Re-sweep against 2.18.1 and land whatever else the diff turns up beyond #419/#420
- [ ] Add a scheduled CI job that diffs the latest `google-genai` release against the last-swept version and opens an issue when `_gaos/types/interactions/**` changes — this is the part that actually prevents recurrence
- [ ] Consider checking in the last-swept version as a machine-readable constant so the job has an unambiguous baseline

## Related: the pinned revision gives no drift signal either

`src/http/common.rs:48` pins `Api-Revision: 2026-05-20`. A live probe sent `Api-Revision: 2099-01-01` and the API returned 200 without complaint — the header is not validated server-side, so there is no failure mode that reveals a newer revision exists. The scheduled sweep is the only realistic detector for both problems.

evansenter:
Closes #421.

`docs/INTERACTIONS_API_GAP.md` had every item struck through under a "Status: implementation tracker" header, with nothing distinguishing "verified complete as of a sweep" from "complete". A reader — human or agent — reasonably concluded the Interactions surface was fully covered. It wasn't: the 2.17.0 → 2.18.1 diff contained `Content::Video.processing` (#419) and a widened `speech_config` union (#420).

Cross-checking `ai.google.dev` would not have caught either — the published reference documents neither. Only diffing the generated bindings surfaced them, and only a live probe confirmed what actually works.

## The verification ordering, now stated in the file and CLAUDE.md

1. **Generated bindings** — leading indicator, ships ahead of prose
2. **Live probes** — ground truth, often narrower than the spec
3. **Prose docs** — lag both; absence is not evidence of absence

Rank 2 earns its place independently: the bindings describe a `speech_config` object arm the API rejects, and `Tool::Retrieval` is spec-present but Vertex-only. New surface must be probed before it's modeled as usable.

`CLAUDE.md`'s external-docs table previously called `ai.google.dev` "authoritative" with no caveat — which is where the wrong conclusion starts for anyone reading this repo cold.

## The sweep job

`.github/workflows/api-surface-sweep.yml` — daily, compares the latest `google-genai` on PyPI against `.github/last-swept-sdk-version` and diffs `_gaos/types/interactions/**`. Two signals: added/removed *files* mean new types; changed field lines mean widened fields.

**Verified against the drift that motivated it.** Replaying 2.17.0 → 2.18.1:

```
files.diff:  > mediaprocessing.py
             > speakerconfig.py
             > staticmediaprocessing.py
full.diff:   SpeechConfigUnion, processing:
```

Both real finds surface, in 9KB — under the 40KB body cap. The no-op path also verified: baseline 2.18.1 vs PyPI latest 2.18.1 exits `changed=false` without downloading.

One rolling issue with a fixed title, same reasoning as #425 — a daily job with a date-stamped title is exactly how 15 duplicate reports accumulated. The issue body tells the reader to live-probe before modeling, and that bumping the baseline closes the loop.

## On the `Api-Revision` half

Left as a note rather than a detector. The API accepted `Api-Revision: 2099-01-01` with a 200, so the header is unvalidated and a new revision has no observable footprint. The SDK sweep is the realistic proxy.

---


evansenter:
## Feedback Addressed

**[Important] Item 21 marks something done that doesn't exist here** — correct, and it's the sharpest possible finding against this PR specifically: I wrote an overclaiming tracker entry inside the change whose purpose is to stop the tracker overclaiming.

**Item 22 was overclaimed too**, which your spot-check didn't catch. You verified it by finding `deserialize_speech_configs` at `src/request.rs:677` — the function exists, but on this branch it's still the old two-arm version:

```rust
enum ListOrSingle {
    List(Vec<SpeechConfig>),
    Single(SpeechConfig),
}
```

No `Speakers` arm. So verifying by symbol name produces a false positive — which I think is a *stronger* argument for the separation than the original finding, since it shows even a careful reviewer can be fooled by a struck-through line plus a plausible symbol.

Both items now sit under an explicit **"found, NOT yet modeled"** heading with `Tracked in #419` / `Tracked in #420`, and the Source paragraph no longer counts them as landed.

**[Important] A moved bindings path reads as no drift** — correct and nasty, because the failure is permanent: the baseline is never bumped in that state, so the job re-reports "unchanged" every morning forever. Both paths are now asserted before the diff with a loud `::error::` and a nonzero exit. Verified the guard fires on a missing directory.

**[Suggestion] `--no-deps` ≠ wheels-only** — applied, `--only-binary=:all:`. My comment claimed something the flag didn't do. Verified `pip download google-genai==2.18.1 --no-deps --only-binary=:all:` still resolves a wheel.

**[Suggestion] triple backticks in upstream docstrings** — applied, tilde fences.

**[Suggestion] dedupe searches open only** — applied, `--state all` plus `gh issue reopen` (no-op when already open). Closing the rolling issue without bumping the baseline would otherwise have forked a duplicate daily, which is the same class as the #441 finding.

---

evansenter:
Round 3 addressed in 5e4bb86. All four findings taken.

**[Critical] line 159 — confirmed and fixed.** Your reading of the tokenization is exactly right, and I reproduced it standalone before touching anything:

```
$ bash -n q.sh
q.sh: line 2: unexpected EOF while looking for matching `"'
exit=2
```

The `'"'"'` idiom is for embedding an apostrophe in a *single*-quoted string; inside a double-quoted one the apostrophe is already literal, so all it does is close the quote and open one that never closes. Now written bare, matching line 152. Your point that the no-op path exits at line 57 before bash ever parses the block is the reason this got through my verification.

**[Suggestion] line 39 — taken, and shipped here rather than deferred.** You're right that nothing in the repo parses workflow shell, and that a schedule-only job's breakage arrives as a cron email rather than a red check. Added `.github/scripts/check_workflow_shell.py`, which pulls every `run:` body out of every workflow and runs `bash -n` over it, wired into `rust.yml` as its own job with no Rust toolchain so it reports in seconds.

I verified it in both directions rather than just that it passes:

```
# on the pre-fix tree
::error file=.github/workflows/api-surface-sweep.yml::sweep / Compare latest SDK
  against the recorded baseline: unexpected EOF while looking for matching `"'
1 run: block(s) failed to parse.  exit=1

# after the fix
All run: blocks parse across 8 workflow file(s).  exit=0
```

I went with `bash -n` over actionlint deliberately: it needs no third-party action or curl-to-bash installer, and it catches this exact class. The docstring is explicit that it is a syntax gate only — it will not catch an unset variable or a wrong flag. If shellcheck-grade linting is wanted across the directory later, that is a bigger change touching all 8 files and belongs on its own.

**[Suggestion] line 139 — taken.** The cap now applies to the assembled `sweep_report.md`, not just `full.diff`. Your files.diff-as-a-double-listing scenario is the realistic trigger and the old cap did nothing for it. On truncation it appends a fence close and `</details>` — a spurious empty fence when the cut landed outside one is cosmetic; an unterminated one swallows the rest of the issue.

**[Suggestion] line 161 — taken.** A new step closes the rolling issue once `changed == 'false'`, which covers both clean paths (baseline caught up, and SDK moved with bindings unchanged). I put the close in the workflow rather than in step 4's prose because a manual step someone can forget is exactly how the header's staleness signal would latch on permanently. The title now lives in one job-level env var used by both the open and close paths — two copies that drifted would leave the close silently matching nothing. `docs/INTERACTIONS_API_GAP.md` header updated to say the close is automatic, so an open issue always means genuinely unswept surface.

---

evansenter:
All five round-4 suggestions applied in b1fede0.

**`check_workflow_shell.py:34` — shell allowlist.** Taken. Now matches the leading word, so `bash -euo pipefail {0}` and any other template stay covered. Verified rather than assumed — a deliberately broken block under exactly that template:

```
::error file=.github/workflows/x.yml::a / (unnamed): unexpected EOF while looking for matching `"'
1 of 1 run: block(s) failed to parse.  exit=1
```

**`check_workflow_shell.py:98` — files vs blocks.** Taken, and this was the sharpest of the five. A gate that goes inert printing the same line as a gate that passes is worse than no gate, precisely because this one covers a class nobody notices until 8am UTC. It now counts blocks and hard-fails on zero:

```
# real tree
All 62 run: blocks parse across 8 workflow file(s).  exit=0

# a workflow with no run: blocks
::error::No run: blocks were checked — the gate is inert.  exit=1
```

**`check_workflow_shell.py:16` — expression substitution.** Taken. The docstring now says outright that `${{ }}` is expanded textually before bash parses, so this checks the template and not the script that runs, with your PR-title-containing-a-quote case as the concrete example. I added the `sh`-parsed-as-bash caveat from your other note in the same block, since both are "this pass is narrower than it looks".

**`api-surface-sweep.yml:278` — the `sed` is a no-op.** Confirmed and removed. I checked rather than taking it on reading, by dumping the parsed `run:` body:

```
"cat > close_comment.md <<'EOF'"
'The Interactions bindings show no diff against the recorded baseline,'
...
'EOF'
```

Column 0, exactly as you said — the block scalar had already stripped it, and my comment had the reasoning backwards. Your point about the latent cost is the part that made it worth deleting rather than leaving: it would have silently eaten ten spaces from any indented line someone added to that comment later.

**`api-surface-sweep.yml:270` — close comment overclaims.** Taken. You're right that the step fires on both clean paths and the text was only true on one. Now "no diff against the recorded baseline", which holds either way, and I applied the same correction to the `docs/INTERACTIONS_API_GAP.md` sentence that had inherited the narrowing.

Also noting your check that `api-sweep` doesn't exist as a label yet and that a labels filter on a nonexistent label returns empty rather than erroring — that's the first-run-after-merge case for the close path specifically, and it's the failure mode I'd have missed. Confirmed safe under `set -euo pipefail`.

---

evansenter:
Round 6 applied in 2ae6310 — all three taken, plus the jq hardening from the same comment.

**Tracker header vs. baseline drift.** This was the best of the three, because the failure mode is a *convincing* wrong state rather than a visible one: bump the baseline without the header and the sweep issue closes, the staleness signal clears, and the header keeps advertising an older snapshot with nothing contradicting it. Added `.github/scripts/check_gap_tracker_baseline.sh`, on the `tests/model_literals.rs` model you pointed at. Verified in both directions rather than only that it passes:

```
Gap tracker and CI baseline agree: google-genai 2.18.1.        exit=0

# baseline moved to 2.99.0
::error file=docs/INTERACTIONS_API_GAP.md::Gap tracker says 2.18.1,
  .github/last-swept-sdk-version says 2.99.0.                  exit=1
```

It also fails rather than passing vacuously if the row it parses ever moves — an empty capture is an error, not a match.

**Truncation notices pointing nowhere.** Correct, and worse than I'd written it: the log only ever receives the already-truncated report, so "read the job log for the full diff" was pointing at the truncated thing. Added an `upload-artifact` step for `full.diff`/`files.diff` on the changed path, and both notices now name the `sweep-diffs` artifact. Your framing is what made it worth doing — the trigger is a wholesale upstream restructure, which is exactly when the untruncated diff matters most.

**`--label api-sweep` is load-bearing.** Taken as a comment rather than a code change. Both keys are needed, but they fail differently, and that asymmetry is the useful part to record: retitling breaks the open and close paths together and loudly, while dropping the label breaks them apart and silently. I said so where the comment previously credited the title alone.

**jq interpolation.** Also taken. Both filters now read `$ENV.SWEEP_ISSUE_TITLE`. Confirmed against jq directly, including the case the old escaped form would have broken on:

```
$ SWEEP_ISSUE_TITLE='has "quotes" in it'   # matches, returns 42
```

`Workflow Shell` now reports 67 blocks across 8 workflows and 4 scripts — the new script is covered by the gate it ships alongside.

---

PR #440, opened 2026-08-16
Run coverage under nextest, and fix the LOUD_WIRE test race it exposed

### LINKED ISSUE #418: Code Coverage job flakes on lenient_vec_warns_are_not_silent (thread-local subscriber vs process-wide interest cache)
`serde_util::tests::lenient_vec_warns_are_not_silent` fails intermittently, and **only** in the Code Coverage job. Every other test job is unaffected, which is the clue to the cause.

## Observed

On #416, commit `103b0de` ([run 31953663811](https://github.com/evansenter/genai-rs/actions/runs/31953663811)):

```
test serde_util::tests::lenient_vec_warns_are_not_silent ... FAILED

thread 'serde_util::tests::lenient_vec_warns_are_not_silent' panicked at src/serde_util.rs:439:9:
the element-drop degradation must warn; got:
  ["List field of i64 was explicit null; degrading to an empty list"]

test result: FAILED. 857 passed; 1 failed
```

The same test passed on `75b7f80` and again on `6a0f03d` with no relevant change between them, so it is genuinely intermittent rather than a regression.

Note the captured vector: the **first** warn (`explicit null`) is present, the **second** (`dropping`) is missing. The test did not fail to capture — it captured, then stopped capturing partway through the closure.

## Why only coverage

| Job | Runner | Isolation |
|-----|--------|-----------|
| Test Suite, strict-unknown, antigravity, integration | `cargo nextest run` | process per test |
| **Code Coverage** | `cargo llvm-cov` → `cargo test` | **threads in one process** |

`test_subscriber::capture` installs its recorder with `tracing::subscriber::with_default`, which is **thread-local**, while tracing's callsite interest cache is **process-wide**. Under nextest nothing else shares the process, so the mismatch is invisible. Under `cargo test` other test threads hit the same callsites concurrently with no subscriber installed.

`src/test_subscriber.rs` already documents this hazard and calls `rebuild_interest_cache()` to mitigate it:

```rust
// Other tests exercise the same callsites with no subscriber
// installed, which can cache their interest as `never`
// process-wide. Under nextest (process per test) that's
// invisible, but under plain `cargo test` (e.g. the coverage
// job) the cache is shared across threads -- rebuild so the
// scoped Recorder is consulted.
tracing::callsite::rebuild_interest_cache();
```

That narrows the window but cannot close it. The rebuild happens once, before `f()`. Any other thread reaching those callsites *during* `f()` can re-evaluate interest with no subscriber present and re-cache `never`, disabling the callsite for the rest of the closure. A failure between the two `vec_roundtrip` calls yields exactly the observed output: first warn captured, second dropped.

## Suggested fix

Point the coverage job at nextest so it gets the same per-process isolation as every other job:

```diff
-        run: cargo llvm-cov --workspace --lcov --output-path lcov.info
+        run: cargo llvm-cov nextest --workspace --lcov --output-path lcov.info
```

`cargo-llvm-cov` supports `nextest` as a first-class subcommand and `cargo-nextest` is already installed in other jobs. This removes the entire class of failure rather than narrowing the window again, and keeps one test runner across CI.

Alternatives, if the coverage job should keep using `cargo test`:

- Serialize the capture-based tests behind a shared `Mutex` so no unsubscribed thread runs those callsites concurrently. Contains it, but every future capture test has to remember to opt in.
- Install a process-global subscriber for the whole test binary and filter per test, so interest never resolves to `never`. More invasive, and changes how all tests observe tracing.

## Impact

Cosmetic but corrosive: coverage is not a required check, so this does not block merges, but it produces red runs unrelated to the change under review, which trains people to ignore the job. It cost a merge cycle on #416.

Surfaced while releasing 0.10.0 (#416).

evansenter:
Closes #418.

The Code Coverage job was the only one using plain `cargo test` (threads in one process); every other job uses nextest (process per test). Several unit tests assume process isolation, so they flake only there.

## Two distinct mechanisms, same class

- **#418's**: `test_subscriber::capture` installs a *thread-local* subscriber while tracing's callsite interest cache is *process-wide*. The existing `rebuild_interest_cache()` narrows the window but cannot close it.
- **LOUD_WIRE**: `Client::builder().build()` reads the env var at construction, so *every* client-building test reads it, not just the one that sets it. A client built inside the set/remove window picks up a stray `LoudWirePrinter`, and `test_add_wire_inspector_accumulates` sees 3 inspectors instead of 2.

## Reproduction

Plain `cargo test --lib` didn't flake in 15 runs, nor `serde_util`-only at 16 threads in 40. It reproduces with the **full** suite at high contention (16 threads on 4 cores) — which fits the analysis: the race needs another thread hitting those callsites during the closure.

**19 failures in 30 runs.** But always a different test than #418 names:

```
11 test client::tests::test_add_wire_inspector_accumulates ... FAILED
```

`lenient_vec_warns_are_not_silent` never failed locally in ~85 runs.

Its `SAFETY` comment asserted the opposite of reality:

> // SAFETY: test-only env mutation. No other test reads LOUD_WIRE, and
> // an extra printer on an unrelated concurrently-built client is harmless

Both halves false — the read is indirect through `build()`, and the extra printer is exactly what breaks the sibling test's count. Under nextest it never showed, so the claim was never tested.

## What landed

1. **Coverage → nextest**, as #418 proposed, plus the `install-action@cargo-nextest` step it needs (matching the spelling the other five jobs use).
2. **The LOUD_WIRE race fixed at the source**, because `cargo test` is a documented local workflow in CLAUDE.md and CI passing isn't a reason to leave a real race. A mutex serializes the tests whose result depends on the variable; lock poisoning is deliberately ignored, since there's no invariant beyond ordering.

**Same stress after the fix: 0 failures in 30 runs.**

I left the tracing/interest-cache mechanism alone — the nextest switch covers it in CI, and #418 weighs the alternatives as worse. It stays reachable for `cargo test --lib` at high parallelism locally; the same mutex treatment would work if that becomes an annoyance.

---


evansenter:
## Feedback Addressed

**[Important] `src/wire.rs` LOUD_WIRE mutator unserialized** — correct, and the fix was incomplete in exactly the way described. `env_inspector_applies_the_filter_from_the_variable` carried the same nextest-per-process SAFETY claim I'd identified as false in the client.rs sibling, and the guard being private to `client.rs`'s `mod tests` meant it couldn't have taken the lock even if it wanted to.

Moved the lock into `src/test_subscriber.rs` — your suggestion, and the right home since it's already the crate's test-support module and reachable from both.

**[Suggestion] `remove_var` clobbers the ambient value** — also correct, and it composes with the above, so I folded both into one RAII type rather than a bare mutex:

```rust
pub(crate) struct LoudWireGuard {
    _lock: MutexGuard<'static, ()>,
    prior: Option<String>,
}
```

`acquire()` takes the lock and records the ambient value; `set()`/`unset()` mutate; `Drop` restores. That makes `LOUD_WIRE=1 cargo test --lib` behave — previously the first LOUD_WIRE test to run would clear it for the rest of the process.

Poisoning is still ignored on acquire, for the reason in the original comment: the lock orders access, it doesn't protect an invariant.

### Verification

Same stress that produced the original 19/30: **0 failures in 65 runs** of `cargo test --lib --test-threads=16`.

One note on my own process — an intermediate run showed 2/30 and I nearly reported the fix as partial. Those two were `cargo fmt` rewriting `src/client.rs` mid-loop in the same command, not races; the 65 clean runs were after the tree settled. Worth stating since "mostly fixed" would have been the wrong claim.

I left the tracing interest-cache mechanism alone, as in round 1 — the nextest switch covers it in CI, and #418 weighs the alternatives as worse.

---

evansenter:
All three round-3 suggestions applied in d88d7f2.

**Duplicated `prior` capture.** Taken — this was the one that mattered. You're right that the test was pinning `nested()`'s capture, so the assertion would have stayed green through exactly the `acquire()` regressions it exists to catch. Both constructors now go through a private `with_lock(Option<MutexGuard>)` that owns the `env::var` read; `acquire()` is `with_lock(Some(..))` and `nested()` is `with_lock(None)`. The reentrancy shape is unchanged, as you suggested.

**`#[must_use]`.** Taken, with the message spelling out the failure mode rather than just "unused":

```rust
#[must_use = "the guard releases the lock and restores LOUD_WIRE when dropped, \
              so discarding it leaves the code that follows unserialized"]
```

**SAFETY scope.** Taken. The comment now says the lock orders *mutators*, and names the residual reader hazard concretely — `reqwest::Client` reading `HTTP_PROXY`/`NO_PROXY` inside a sibling `Client` build, glibc potentially freeing an `environ` entry a concurrent `getenv` is mid-read, which is the UB that made `set_var` unsafe in 2024. It also records why it's a note and not a fix: this change narrows the window rather than widening it, and nextest's process-per-test confines it to plain `cargo test`. That last part is the useful half — without it the next reader has to re-derive whether the gap is known or overlooked.

Verification: `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, and 859 lib tests all green.

---

PR #442, opened 2026-08-16
Add a fresh-consumer-crate check for #[tool] requirements

### LINKED ISSUE #402: Macro: emit fully-qualified paths so #[tool] needs no consumer-side imports
Follow-up from #401 (fix 4). Dogfooding the published 0.8.0 showed `#[tool]` fails to compile in a fresh consumer crate unless the consumer:

1. adds `async-trait` and `serde_json` as direct dependencies (the generated code references `::async_trait::async_trait` and `::serde_json::` from the consumer's dependency graph), and
2. has `use genai_rs::CallableFunction;` in scope wherever the macro is applied (`codegen.rs` emits `.declaration()` in method position, and the generated `<fn_name>_declaration()` free function inherits the same requirement).

The docs now state these requirements (README install snippet groups the three deps as "required together"), but the better fix is removing them:

- Re-export `async_trait` and `serde_json` under `genai_rs::__private` (or similar) and have `genai-rs-macros` emit paths through the re-exports, so consumers need only `genai-rs` + `genai-rs-macros`.
- Emit `<genai_rs::CallableFunction>::declaration(&...)` (fully-qualified call syntax) instead of method position, so no trait import is needed at expansion sites.

Once done, revert the README dependency grouping and the "import required at every call site" note, and simplify the ~15 doc snippets that carry the imports. A trybuild fixture compiling a `#[tool]` fn with *no* imports would pin the behavior.

### LINKED ISSUE #404: CI: scheduled fresh-consumer-crate compile check for #[tool] requirements
Follow-up from #401 (fix 4). The `#[tool]` consumer requirements (deps + trait import — see #402) can only fail from a *fresh consumer crate*: inside this workspace `async-trait`/`serde_json` are always present and `CallableFunction` always resolvable, so neither doctests nor trybuild fixtures can reproduce the failure. It took manual dogfooding of the published 0.8.0 to find, and manual dogfooding is the only thing that would find it again if `codegen.rs`'s emitted paths change.

Proposal: a scheduled (not per-PR) job that scaffolds a throwaway crate against the path dependency, copies in the README's exact Installation snippet + Basic Usage example, and compiles one `#[tool]` function. That exercises exactly the dependency graph the README documents.

Superseded if #402 lands and removes the requirements outright — in that case the job instead pins that a consumer with *only* `genai-rs` + `genai-rs-macros` compiles.

evansenter:
Closes #404. Reopened #402 on the strength of what this found.

The consumer requirements `#[tool]` imposes can only fail from outside this workspace: in-tree, `async-trait` and `serde_json` are always present and `CallableFunction` is always resolvable, so no doctest or trybuild fixture can reproduce them. Finding the 0.8.0 breakage took manual dogfooding of the published crate. This is that dogfooding, automated.

## Two jobs, two questions

**`documented-requirements`** builds and runs a crate using README.md's exact Installation dependency set plus the documented `use genai_rs::CallableFunction;`. This is the gate: it fails if `codegen.rs`'s emitted paths drift from what the README tells users to do. Verified locally against the identical scaffold.

**`minimal-requirements`** builds a crate with only `genai-rs` + `genai-rs-macros`. Expected to fail while #402 is open, so it's `continue-on-error` and reports rather than gates. When it starts passing, #402's requirements are actually gone and it should become a gate.

## Why that second job exists

I closed #402 earlier today on a reading of `codegen.rs` — it emits `::genai_rs::…` paths, so it looked done. Building the crate proved otherwise:

```
error[E0433]: failed to resolve: could not find `async_trait` in the list of imported crates
error[E0433]: failed to resolve: could not find `serde_json` in the list of imported crates
```

then, after adding both:

```
error[E0599]: no method named `declaration` found for struct `GetWeatherCallable`
help: trait `CallableFunction` which provides `declaration` is implemented but not in scope
```

`::serde_json::Value` **is** a fully-qualified path — the `::` means no `use` is needed — but it still resolves against the *consumer's* dependency graph, so the crate must be a dependency regardless. Fully-qualified ≠ dependency-free, and I conflated them. A compiler answers that question correctly; grep does not. #402 is reopened with the three precise sites.

Path dependencies rather than the published crate, so an unreleased codegen change fails here rather than after a release.

Scheduled weekly rather than per-PR — it guards against codegen drift, not the average change — plus on PRs touching `genai-rs-macros/src` or `README.md`, the two inputs it verifies.

---


evansenter:
Disposition on the six suggestions:

**Will fix:**

- **`if: failure()` is over-broad.** This is the one worth fixing regardless of anything else — a flaky apt mirror emitting `A fresh consumer crate failed to compile a #[tool] function` is the most misleading possible annotation on a job whose entire value is that its red means one specific thing. Scoping to `steps.build.conclusion == 'failure'`.
- **Stale line references.** `README lines 99-105` not `99-104`, and naming anchors rather than line numbers so both survive edits.
- **The comment overstates what `main()` exercises.** Adding the factory assertion you sketched — it's one line and it exercises the `Box<dyn CallableFunction>` construction and trait-object dispatch the registry actually uses, so the comment becomes true rather than being softened.
- **mold installed but unused.** Correct — cargo config discovery walks up from `consumer/`, never sees `genai-rs/.cargo/config.toml`. Dropping the step rather than adding `RUSTFLAGS`: this job's cost is dominated by compiling the dependency tree from scratch, and ~30s of apt to speed up a weekly run isn't worth the extra moving part.

**Won't fix here, with reasons:**

- **Reading the dependency set from README.md.** You've identified the real blind spot precisely: someone drops `async-trait` from the snippet believing #402 is fixed, this job triggers on `README.md`, still hardcodes it, and goes green. That's a genuine hole. But extracting and rewriting a toml block out of markdown in shell is fragile in a way that fails *open* — a parse that silently yields an empty dependency list gives a green job testing nothing, which is worse than the hole it closes. The `minimal-requirements` job already covers the specific scenario you name from the other direction: if #402 lands and the requirements really are gone, that job goes green and tells us. I'd rather close this by making the minimal job a gate once #402 lands than by parsing markdown.
- **`rust-cache` workspaces.** Real, and your diagnosis is right that artifacts land in `consumer/target`. For a weekly job I'd rather have the from-scratch build: it's closer to what a new consumer actually experiences, which is the whole premise. Worth a comment saying that's deliberate rather than leaving it looking like an oversight — will add one.

---

Issue #428, opened 2026-08-16
Checked-in `.cargo/config.toml` hard-requires mold; a clone without it fails with a cryptic `cannot find 'ld'`


evansenter:
`.cargo/config.toml` unconditionally sets the mold linker for every build:

```toml
[target.x86_64-unknown-linux-gnu]
rustflags = ["-C", "link-arg=-fuse-ld=mold"]
```

On a machine without mold installed, every build fails before compiling anything:

```
error: linking with `cc` failed: exit status: 1
  = note: collect2: fatal error: cannot find 'ld'
  compilation terminated.
error: could not compile `proc-macro2` (build script) due to 1 previous error
```

The message is actively misleading — it says `ld` is missing when `ld`, `lld`, and `gold` are all present and the actual missing binary is `mold`. Nothing in the error names mold, so the trail back to `.cargo/config.toml` is not obvious. The file's own comment (`# Install mold: sudo apt install mold`) is the only pointer, and you have to already suspect the linker to go looking for it.

CI installs mold explicitly, so this only bites new contributors, fresh containers, and ephemeral dev environments — which is exactly where a cryptic first-build failure is most expensive. Hit while working #419 in a fresh container; worked around with `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS="-C link-arg=-fuse-ld=lld"`.

## Options

1. **Move it to `.cargo/config.toml.example`** and document opting in. Loses the default speed-up for everyone who does have mold.
2. **Keep the default, make the failure legible.** A `SessionStart`/setup script or a README line that names mold explicitly. Cheapest, but still fails first and explains second.
3. **Prefer mold, fall back.** Cargo config has no conditional-on-binary-present mechanism, so this needs a wrapper or a build script — probably more machinery than the problem justifies.

Leaning 1 or 2. Worth noting #342 introduced this for build speed, so the intent is sound; only the unconditional-ness is the problem.

PR #443, opened 2026-08-16
Replace the absolute example-size ceiling with a baseline-delta check

### LINKED ISSUE #403: CI: replace the absolute example-size ceiling with a baseline-delta check
Follow-up from #401 (fix 1). The Build Metrics size gate is an absolute ceiling that has now been raised three times (10 → 12 → 16 MB), each time only after it went red on an unrelated PR. The failure mode is structural: the tripwire fires once drift has accumulated past the margin, and the remedy is always to widen the margin — so it catches toolchain drift loudly while a genuine per-PR regression under the ceiling passes silently.

The Build Metrics job already measures per-example sizes. Proposal:

- Commit a `size-baseline.json` (or read the previous run's artifact) with per-example release sizes.
- Fail on a **delta** — e.g. >15% growth for a single example within one PR — instead of an absolute number.
- A scheduled job (or the release workflow) refreshes the baseline so routine toolchain drift is absorbed without a manual bump.

This catches the regression on the PR that introduces it, and stops the every-couple-releases ceiling bump.

evansenter:
Closes #403.

The ceiling had been raised three times (10 → 12 → 16 MB), each time only after going red on an unrelated PR. That failure mode is structural: an absolute tripwire fires once drift accumulates past the margin, and the only remedy is widening the margin — so it caught toolchain drift loudly while a genuine per-PR regression *under* the ceiling passed silently. A delta check inverts both halves.

Fails when any example present in both baseline and current grows more than 15%. New examples are reported but never fail — nothing to compare against, and a first appearance isn't a regression.

## Artifact, not a committed file

Took the artifact option for the reason the issue itself raises: sizes are toolchain- and linker-dependent, so a committed number encodes one machine's toolchain and drifts against every other — the same shape as the ceiling problem, relocated. The artifact refreshes on every main push, so routine drift is absorbed with no manual step.

## Tested

Logic lives in `.github/scripts/check_example_sizes.sh` rather than inline YAML so it can be tested, and was:

| Case | Result |
|---|---|
| within threshold | pass |
| one example +30% | **fail**, per-example `::error::` |
| no baseline | pass, with a notice |
| shrinkage (−50%) | pass |
| `measure` with `.d` files and hash-suffixed duplicates | correctly skipped |

That last one matters more than it looks: cargo leaves both `simple_interaction` and `simple_interaction-a1b2c3…` in the examples dir. Without filtering, the key set changes every rebuild and every entry reads as "new", silently disabling the check.

## ⚠️ This changes when a job runs

Build Metrics was `if: github.event_name == 'pull_request'` — PR-only. Fine for an absolute ceiling, but a delta check has nothing to compare against, and the baseline-publish step would never have run. The trigger now includes pushes to main, **roughly doubling how often this job runs**. That's the real price of the automatic-refresh property, and the design point here I'd most want a second opinion on.

Also added `permissions: {contents: read, actions: read}` — `gh run download` needs `actions: read`, and the workflow had no `permissions` block at all.

## Kept

A 64MB sanity ceiling, deliberately far above anything plausible, purely for the no-baseline case. The delta check is the gate; that's a backstop. Threshold is the script's third argument, so it's a one-word change.

---


evansenter:
Working through these. Disposition up front so nothing sits ambiguous:

**Will fix — these are real defects, not polish:**

- **Zero comparable entries reports success.** You're right that the PR description identifies this as the dangerous mode and then doesn't guard the symptom, only one known cause. Adding a comparable-count check that fails (or warns) when it's 0 while both files are non-empty, plus the same for `measure` finding 0 binaries.
- **`over` compares unrounded, prints rounded.** `+15.04%` failing with the message `grew +15% ... over the +15% threshold` reads as a checker bug. Will keep the raw value for the comparison and print an extra decimal.
- **The escape hatch doesn't exist.** "Say so in the PR" describes nothing — no workflow reads the body or labels. Rewording to name the two routes that actually exist rather than implementing a label hatch, since a label is a bigger decision than this PR should make unilaterally.
- **`.[0].databaseId` yields `null`, not empty.** Good catch — `[ -n "null" ]` passes and we then `gh run download null`. Appending `// empty`.

**Considering, not yet decided:**

- **`--limit 1` on the baseline lookup.** Walking back a few runs is clearly more robust. Weighing it against the failure being visible rather than silent — which your other suggestion (`::warning::` instead of `::notice::` on the skip path) addresses more directly and more cheaply. Likely doing the `::warning::` change and leaving the single-run lookup.
- **Artifact expiry over a quiet period.** This is the sharpest structural point in the review, and the 2026-03-29 → 2026-07-08 gap in this repo's own history makes it concrete rather than hypothetical. The +29% toolchain drift you cite would trip the threshold on the first PR after a quiet spell — reintroducing the false-positive class #403 set out to remove. I don't think I should pick between `workflow_dispatch` republish and max-of-last-N inside this PR; it deserves its own issue with the history quoted.

**Not doing here:**

- **Shellcheck + a bats-style harness for `.github/scripts/`.** Agreed on the substance — the matrix in the PR description is exactly the kind of thing that decays to nothing once it's only in prose. But it covers four scripts, three of which this PR doesn't touch, so it belongs in its own change rather than riding along.

---

evansenter:
Round 10 suggestion addressed in 1e0c062.

The baseline fetch now branches on `gh run list` failing rather than folding it into the empty case:

```bash
if ! RUN_IDS=$(gh run list --workflow Rust --branch main --status completed \
    --limit 10 --json databaseId --jq '.[].databaseId // empty' 2>/dev/null); then
  echo "::warning::Could not list main runs — the baseline lookup failed rather than finding nothing. Check the actions:read permission and rate limits."
  RUN_IDS=""
fi
```

Behaviour is unchanged — both paths still land in the same non-fatal `No usable baseline` branch, which is honest since the delta check genuinely cannot run either way. What changes is that the log now says which one happened. That matters because the fallback gate is only the 64MB backstop: a missing artifact self-clears on the next main push, but a dropped `actions: read` scope or a rate-limited token does not, and previously rendered identically to the benign first-PR case on every run.

Simulated all three paths against the edited block: `gh` erroring takes the warning branch with `RUN_IDS=''`; zero runs found stays silent with `RUN_IDS=''`; runs found yields the ID list. All `build-metrics` run blocks still pass `bash -n` and the workflow parses.

---

evansenter:
Round 11 addressed in 333d3f1. Both suggestions taken; on the first I took the "cover them" branch rather than the "say so" one.

**1. antigravity examples are now built and measured.** The finding is right and the comment was claiming coverage that never existed — `required-features = ["antigravity"]` on all seven targets, `antigravity` not in `default`, and Cargo silently skips a target whose required features are off. `Build examples` now runs:

```
cargo build --release --examples --features antigravity
```

Covering them beat documenting the gap for the reason you gave: it is the largest feature surface in the crate, so it is where a size regression is least likely to be noticed by any other means. Two things that make it cheap — they already compile in CI (`check` runs `--all-targets --all-features`), so this changes only what gets linked and measured; and although enabling the feature shifts every *other* example's size too, baseline and PR are measured by this same step, so both sides move together and the deltas stay meaningful. They will show up once as `::notice::No usable baseline` and be gated from the next main push onward, as you predicted.

**2. Delta table now goes to the step summary too**, with all three caveats applied:

```bash
set -o pipefail
{ echo '### Example size deltas vs main'; echo ''; echo '```'; } >> $GITHUB_STEP_SUMMARY
rc=0
.github/scripts/check_example_sizes.sh compare \
  baseline-sizes.json current-sizes.json 15 \
  | tee -a $GITHUB_STEP_SUMMARY || rc=$?
echo '```' >> $GITHUB_STEP_SUMMARY
exit "$rc"
```

The closing fence is unconditional rather than inside the success path — an unterminated fence would swallow whatever a later step appends to the summary.

Verified by simulating the block with `compare` stubbed to exit 0 and then 1:

```
compare rc=0 -> step exit 0    fences: 2
compare rc=1 -> step exit 1    fences: 2
```

so the gate survives the pipe, and the table reaches the summary on both paths. YAML parses and every `run:` block in the file passes `bash -n`.

---

evansenter:
Round 12 addressed in 92815c3. All three taken.

**1. `Clean build time` builds with `--features antigravity` too.** You're right that this was a real cost, not just a cosmetic mismatch — the `Examples build` row was absorbing a full lib rebuild plus `tokio-tungstenite`/`tungstenite`/`libc`, which would step up once at merge and then hide any genuine change in example build time inside it. Both rows now describe the same build, and the lib is compiled once per run rather than twice, which matters more now that the job also runs on merges.

**2. `2>/dev/null` dropped from the `gh run list` lookup.** The warning tells the reader to check the `actions: read` permission and rate limits, and gh's own message is the thing that distinguishes those — suppressing it sent them looking for something the log had already discarded. The message now says "See the gh error above". The `gh run download` below keeps its suppression, where a missing artifact really is the expected case. `RUN_IDS` captures stdout only, so nothing contaminates the id list.

**3. `make test-scripts`**, and CI now calls that rather than naming the script:

```make
test-scripts:
	@found=0; \
	for t in .github/scripts/tests/*.sh; do \
		[ -e "$$t" ] || continue; \
		found=1; echo "==> $$t"; "$$t" || exit 1; \
	done; \
	[ "$$found" -eq 1 ] || echo "no script harnesses in .github/scripts/tests/"
```

Globbed deliberately: three other in-flight branches add their own harnesses to that directory, and this way each is picked up by dropping the file in rather than by a second edit to the workflow that is easy to forget. Folded into `make check`. Verified it runs all 43 checks in about a second, and that the empty-directory path is a clean no-op with a message rather than a glob-literal error — which is the state the target starts in on any branch that doesn't have a harness yet.

That also gives the deferred shellcheck-plus-harness sweep somewhere to land.

`rust.yml` parses and every `run:` block passes `bash -n`.

---

evansenter:
Round 13 addressed in 0e92112. All three taken, plus the parenthetical.

**1. `bash "$t"` instead of `"$t"`.** Right — the comment claimed drop-in pickup and the implementation quietly also required `chmod +x`, with a failure that reads as a failing test rather than a missing mode bit. Verified a non-executable harness now runs.

**The parenthetical on the empty glob — taken too, and I went the `exit 1` way.** You're right that "runs nothing, reports success" is the same shape the fixture harness exists to catch in `compare`, and I don't think an empty directory is a legitimate state here: every branch that carries this target also carries at least one harness, so an empty match means the files went missing rather than that nothing is under test yet. It now fails with a message naming the glob. Verified both paths.

**2. `CLAUDE.md` documents the target.** Fair point about it being the contract a future session reads first. Updated the command table (`make test-scripts # Fixture tests for .github/scripts/ (needs only bash + jq)`), the `check` comment in that table, and the quality-checks section further down, which had the same `fmt + clippy + test` phrasing.

**3. `--workflow rust.yml`.** Taken for the reason given — the display `name:` is the cosmetic identifier, so it's the one that gets edited by someone with no reason to look at this job, and the failure branch below only makes the rename *visible*, after every PR in the meantime has already run on the 64MB backstop alone. Same reasoning as pinning the hash-suffix filter rather than relying on downstream diagnostics.

`make test-scripts` passes (43 checks), the empty-directory and non-exec paths both behave as described, and `rust.yml` parses.

---

evansenter:
Round 14 addressed in 1586aff. All five inline items across the two reviews taken.

**1. Baseline guard now checks shape, not just parseability.**

```bash
if [ ! -s "$baseline" ] || ! jq -e 'type == "object"' "$baseline" >/dev/null 2>&1; then
```

Fixtures added for `[]`, `"a string"`, `null` and `42` — all four now land in the honest `::warning::No usable baseline` branch rather than aborting on `Cannot index array with string`. Mutation-tested: reverting to `jq empty` fails the harness. Agreed it's shape-drift insurance rather than a live path while `measure` writes every baseline, but the failure it prevents is the same broken-checker-reads-as-regression confusion the neighbouring guard exists for.

**2. `make test-scripts` no longer stops at the first failing harness.** Records `rc=1` and defers, so a change touching two scripts reports both. Exit status identical; verified with two deliberately-failing harnesses that both print and the target still exits non-zero.

**3. `jq` preflight, in both the harness and the make target.** This one is the sharpest of the three, for the reason you spell out: it isn't that the failures are unhelpful, it's that they're *wrong*. `! jq ...` succeeds when jq is absent, so every `compare` case falls into the no-usable-baseline branch and the rc-1 cases report a confident wrong verdict — the harness doing precisely what it exists to catch the script doing. Now one accurate message naming jq.

**4. `CLAUDE.md` gains an "Example size gate" section.** Covers the two things you flagged: that `build-metrics` is the one job now running on merges as well and why (the baseline is an artifact refreshed on main pushes; a committed number would encode one machine's toolchain), and the `size-growth-ok` label — including the non-obvious part, that a re-run does not pick it up because the re-run replays the original event payload, so the label must be followed by a push. That caveat was discoverable only from the failure message, i.e. only to someone already looking at a red run.

`make test-scripts` passes (55 checks now), both scripts pass `bash -n`.

---

evansenter:
Round 16 addressed in 2290e70. Both taken.

**1. Threshold validation.** The quoted-string case is the one that made this worth doing — jq orders every number before every string, so `"15"` makes `> $max` false for every example and the job prints `All examples within +"15"% of baseline` and goes green. A gate reporting clean while comparing nothing, which is the disjoint-key failure arriving through a different door. Guarded with a `case` next to the `local` line, with fixtures over `"15"`, `15%`, `-5`, `abc` and `1.5`.

One thing worth flagging: my first cut of those fixtures compared the **over**-threshold pair, and passed for the wrong reason — that pair exits 1 on size regardless of what the threshold says. Retargeted at the under-threshold pair, where a valid threshold exits 0, so a non-zero exit can only be the guard; plus a companion assertion that the same pair still passes with a valid threshold. I also left an empty third argument out of the malformed list, since `${3:-15}` defaults it — that's a legitimate call, not a bad one.

**2. The one-sided input guard — taken, and as a hard error rather than by widening the existing branch.** Your framing is the reason: a bad baseline is benign (nothing to compare against, so warn-and-skip is honest), while a bad current file means the build under test is unmeasurable, and skipping *that* is exactly the vacuous pass the disjoint guard exists to reject. So they get different handling, and the comment records why so the next reader doesn't collapse them. Fixtures over `[]`, `not json` and empty.

Both guards mutation-tested — removing either fails the harness. `make test-scripts` passes (77 checks now), both scripts pass `bash -n`.

---

evansenter:
Round 17 addressed in f0797da. All three taken.

**1. The hash filter now requires the whole tail to be hex.** A bare glob can't express "all hex to end of name", so it's a split `case`: at least 8 characters after the final hyphen, and no non-hex character among them. Verified against a directory containing `plain_name`, `real-deadbeef_suffix`, `short-abc`, `keep-0123456g`, `drop-0123456789abcdef` and `drop-0123456789` — only the last two are dropped, and that's now a fixture.

Your framing of why it matters is what made me tighten rather than shrug: a false positive drops the example from *both* sides, so it's absent from the table, from `added` and from `removed` alike — the compared count is its only trace, and that's precisely the signal the removed/renamed notice can't see.

**2. Both denominators in the success line.**

```
All examples within +15% of baseline (1 compared of 1 built / 3 in baseline).
```

Exactly the partial-divergence case you describe — fatal to coverage, previously indistinguishable from a clean run, and nobody reading `1 compared` remembers it used to be `20`. Fixtured with a 3-key baseline against a 1-key current.

**3. shellcheck, skip-if-absent.** Wired into `make test-scripts` with the shape you suggested, so CI picks it up via the existing step and contributors don't gain a second hard dependency.

I then actually ran it, which is the part worth reporting: **it wasn't clean**, so shipping the gate as-is would have reddened CI. Two findings, both fixed:

- **SC2001** ×3 in the harness — `sed 's/^/         /'` reindents replaced with parameter expansion. Verified the output is byte-identical by forcing a real assertion failure and comparing the indented block.
- **SC2064** in `compare_trends.sh` — the EXIT trap was double-quoted, so the paths were baked in at definition rather than resolved when it fires. No behavioural difference today (all three are set above and never reassigned), but it's a real latent bug and this PR doesn't otherwise touch that file. Fixed rather than suppressed, because a lint gate that starts with known findings isn't a gate.

`shellcheck` now exits 0 across all six scripts. `make test-scripts` passes 81 checks, `rust.yml` parses.

---

evansenter:
Round 18 addressed in 0a31125. Both taken.

**1. `measure` no longer builds JSON by hand.** Went with the `jq -Rn` fold you suggested — name and size as two raw lines per file, folded into the object:

```bash
printf '%s\n%d\n' "$name" "$size"
...
} | jq -Rn '
    [inputs] as $lines
    | reduce range(0; ($lines | length); 2) as $i
        ({}; .[$lines[$i]] = ($lines[$i + 1] | tonumber))
' > "$out"
```

which also drops the comma bookkeeping and the `first` flag. Verified against a directory of adversarial names — `has"quote`, `has\backslash`, `tab<TAB>name`, `ünïcødé`, `plain` — all five come back correctly escaped. One detail worth noting: `%d` rather than `%s` for the size, because BSD `wc -c` pads with leading spaces that `tonumber` rejects. The remaining assumption is that a name contains no newline, which would desynchronise the pairing; cargo target names can't, and that's written down.

You're right that it was unreachable today. The reason I closed it rather than noting it is the one you gave — the script already carries this exact insurance for the threshold and for both input shapes, so leaving the one remaining hand-built JSON as the exception would be inconsistent.

**2. `shellcheck -S warning`.** This is a sharp observation and I'd missed it: an unpinned linter at default severity is a tripwire that fires on toolchain drift rather than on anything the author did, with "loosen it after the fact" as the only remedy — which is *exactly* the failure mode this PR exists to remove. Reintroducing it in the same change would have been unfortunate.

`-S warning` drops the info/style tier where new checks mostly land. Verified the floor still catches real defects: re-breaking the `compare_trends.sh` trap (the SC2064 this PR fixed) is still reported at `-S warning`. Chose the severity floor over pinning, since pinning brings back the second-dependency cost the skip-if-absent design deliberately avoids.

`make test-scripts` passes 84 checks with shellcheck clean; `bash -n` clean.

---

evansenter:
Round 19 addressed in afc9c7d. Both taken.

**1. The escaping had no fixture — and you quoted the right line back at me.** My verification lived in a commit message, which is exactly the "matrix nobody re-runs" this harness's own header exists to argue against. Now a fixture over the four names the old `printf` path got wrong (`has"quote`, `has\backslash`, a tab-bearing name, `ünïcødé`) plus a plain control, asserting each key survives into the map intact rather than just that `measure` exits 0.

The stale seam comment is fixed too — it justified itself by pointing at a hand-built emitter that no longer exists. It now names the two changes that have actually moved those bytes (`stat -c%s` → `wc -c`, and printf → jq), neither of which the surrounding cases would have caught, which keeps the case's reason for existing accurate.

**2. The shape guard checked the container but not the values.** Confirmed the chain you describe: `{"alpha": "1000000"}` is an object, jq orders numbers before strings so `"1000000" > 0` is true, it survives the divisor guard, and then dies at `(.value - $was)` with `string and number cannot be subtracted` — jq exit 5, `report=` down under `set -e`, step red on a raw error. Same broken-checker-reads-as-a-regression failure as the cases the guard already covers, just one filter further along.

Both guards are now `jq -e 'type == "object" and all(.[]; type == "number")'`, and both fixture loops gain the string-size case. Mutation-tested: reverting to the bare `type == "object"` fails the harness.

Also relabelled those loops from "non-object" to "unusable" — an object of strings made the old label a contradiction in terms.

`make test-scripts` passes 96 checks, shellcheck clean at `-S warning`, `bash -n` clean.

---

evansenter:
Round 20 addressed in 2fafff4, which also merges main now that #449 has landed.

**1. The feature-set transition.** Correct that "Harmless" was the word that would send someone hunting for a regression that isn't there. The comment now names the transition explicitly: the baseline was measured at whatever feature set the step had on the last main push, so the PR that *changes* that list compares new-set sizes against old-set ones — and a change this size shifts every example at once, so the whole table trips with nothing in it pointing at the flag as the cause. Named the `size-growth-ok` remedy in the same place, since the person hitting it is reading exactly that comment.

**2. The SC2317 directive.** Right — it's inert under `-S warning`, so it read as a required suppression when it's insurance for a bare `shellcheck` invocation this repo doesn't make. Reworded to say which it is, same class as the stale seam comment last round.

**Merge with main was clean**, and one thing is worth reporting because it exercises a claim this PR makes rather than just asserting it: #449 landed two harnesses into `.github/scripts/tests/`, and `make test-scripts` picked them both up by glob with no edit here — the drop-in property the Makefile comment claims, tested by accident:

```
==> shellcheck
==> .github/scripts/tests/test_check_example_sizes.sh   All checks passed.
==> .github/scripts/tests/test_scheduled_failure_scripts.sh   All checks passed.
==> .github/scripts/tests/test_setup_dev.sh   All checks passed.
```

Both of #449's scripts also pass `shellcheck -S warning`, so the gate this PR adds starts green across the larger script set rather than needing a cleanup pass. `build-metrics` also picked up #449's `setup-mold` composite action cleanly. All four workflows parse and every `run:` block passes `bash -n`.

---

evansenter:
Round 21 addressed in 5b5e00b. All three taken — and these are the most valuable findings of the twenty-one rounds, because all three are *consequences of the merge* rather than anything visible when the PR was written.

**1. `+-0%` — reproduced, and it's the common case, not an edge one.**

```
$ jq: (13000000 → 12999999)  pct=-0  rendered=+-0%
```

`round` on a small negative gives `-0`, which is `>= 0` in jq, so the `+` prefix applies and jq stringifies it as `-0`. Fixed with `+ 0`, applied to **both** the printed and the compared value so they can't diverge — the same trap the +15.04% rounding fix was about. Verified: a one-byte shrink on 13MB now renders `+0%`.

**2. Two entry points, folded into one.** You're right that the "second edit that is easy to forget" comment had come to describe an edit that already existed. Now a single `shell-scripts` job calling `make test-scripts`, and the `build-metrics` step is gone. Your fast-feedback point is the better half of the argument: a shell quoting slip surfacing from the job that opens with `cargo clean` and a full release build is exactly wrong.

To your unverified note — `shellcheck -S warning` does now cover `report_scheduled_failure.sh` and `resolve_scheduled_failure.sh` for the first time, and **both pass clean**, so the consolidation doesn't need a cleanup pass.

**3. The dependency surface — and the third bullet is the one that matters.** Verified all of it:

- `test_setup_dev.sh` shells out to `python3` for the TOML parse (line 141). Now preflighted alongside jq; confirmed a `PATH` without python3 prints `python3 is required by these harnesses` instead of a mystery assertion failure.
- It resolves the real `cc` to probe `-fuse-ld=mold`, so "about a second" was wrong. Comment corrected.
- **It deletes and rewrites the checkout's real `.cargo/config.toml`**, restoring from a temp copy on an `EXIT` trap. That was CI-only when #449 landed it; `make check` running it is *my* change, so this is my regression to own. The sharp edge is that the file is gitignored — that being the entire point of #428 — so git can't bring it back if the trap never fires. Documented at the target and filed as **#455**, with the fix being a target-directory override on `setup-dev.sh` so the harness can work in a temp tree and drop the save/restore dance entirely.

I've left the harness running in `make check` rather than pulling it: the trap covers the ordinary cases including SIGINT, and the coverage is worth more than the residual risk. But it's recorded rather than silent now.

CLAUDE.md's CI job list picks up `shell-scripts`. `make test-scripts` green across all three harnesses, shellcheck clean, `rust.yml` parses with every `run:` block passing `bash -n`.

---

evansenter:
Round 22 addressed in fd53778. Both taken.

**1. The naming assumptions belong to different halves — and my fixture pinned the wrong contract.** The comment said "no newline", which is true of `measure` alone; `compare`'s report line is tab-delimited, so a tab-bearing name splits across `name` and `was` and shifts every column. The escaping fixture deliberately includes a tab-bearing name and asserts it survives, so the harness was claiming support the other half of the script can't deliver.

Fixed both ends: the assumptions comment now names newline *and* tab, attributing each to the half that has it, and the fixture states it scopes to `measure`'s JSON escaping with a pointer to why `compare` couldn't carry that name across the seam. Kept the tab case rather than dropping it — escaping is genuinely what it tests — but it no longer reads as an end-to-end guarantee.

**2. `compare`'s `-ge 2`.** Right that the optional threshold explains two-or-three but not four-or-more. Now `{ [ $# -eq 2 ] || [ $# -eq 3 ]; } || usage`, verified:

```
2 args -> 0     3 args -> 0     4 args (`... 15 --strict`) -> 2 (usage)
```

Previously that fourth argument was discarded silently, so a mistyped flag ran as if it had never been passed.

`make test-scripts` green, shellcheck clean.

---

evansenter:
All three round-24 items applied in `8dc37ac`.

**The `return`-in-a-pipeline note was not just a durability note.** I measured the coupling rather than reasoning about it, and it is worse than the comment claimed:

```
=== tab LAST in glob order, pipefail stripped ===
::error::Example name contains a tab, which the size report cannot carry: zzz	name
Measured 2 example binaries -> /tmp/tt2/out.json
rc=0 ; json: {"aaa": 0, "bbb": 0}
```

Exit **0**, with a truncated baseline published. Every example past the tab is dropped, and — by the argument already in the hash-filter comment — a dropped example appears in neither `added` nor `removed`, so the only trace is the compared count. The guard silently becomes the data-loss bug it was added to prevent. It only ever looked safe in my earlier fixture because the tab name sorted *first*, so the pipeline emitted nothing and the "Measured 0" guard failed the run for an unrelated reason. That fixture was passing for the wrong reason; both new ones sort the bad name last.

Fixed by moving validation to its own pass ahead of the pipeline. `return 1` now returns from `measure`, and `$out` is never written — which the fixtures also assert, since rejecting the run while leaving a partial baseline on disk is the half-done state worth ruling out.

**Newline is now guarded rather than documented.** Your trap warning was correct and I verified it before writing anything: `$(printf '\n')` is a zero-length string, so `*"$(printf '\n')"*` collapses to `**`. Mutating the guard into that form fails 7 assertions — it rejects every name in the directory. Written as `$'\n'` instead. I did not take the `*[[:print:]]*` form: in the C locale that rejects the `ünïcødé` name in the escaping fixture, and non-ASCII names are legal here.

**`render` added to the header usage block.**

Each change mutation-tested — dropping the validation pass, restoring the in-pipeline form, and using the `$(printf)` form all fail the harness:

```
mutation A (no pass):        tab exits 0, newline exits 5 (raw jq parse error) — 6 assertions fail
mutation B (in-pipeline):    partial output file left behind — 1 assertion fails
mutation C ($(printf) form): 7 assertions fail
```

---

evansenter:
All three round-25 suggestions applied in `99c1248`.

**The filter-set drift was the one with teeth.** You framed it as symmetry rather than a defect, but it was reachable without any exotic name: the validation pass rejected on *any* regular file, so a non-executable `notes<TAB>file.txt` sitting in the examples dir failed `measure` outright — over a name the measuring loop would have dropped before it reached anything. Hoisted both loops onto a shared `is_example_binary`, which is the version of the fix that also removes the drift rather than just re-aligning it once.

Fixture pins the agreement, mutation-tested against both directions it can break:

```
mutation: validation reverts to [ -f ]   → stray file: expected exit 0, got 1
mutation: predicate drops -x             → stray file: expected exit 0, got 1
```

**`scripts/` added to the shellcheck glob.** Verified it is already clean at `-S warning`, so this widens coverage without changing the current verdict — worth stating since the point of `-S warning` is to keep this job from reddening on things the author did not touch. Your reason for it is the sharp one: `setup-dev.sh` is the script whose `command -v mold` guard was the original "valid shell meaning the wrong thing", and it had a harness but no lint.

One note on the glob: if `scripts/*.sh` ever matches nothing, the literal pattern reaches shellcheck and the job fails rather than silently linting less. That is the same failure direction as the "an empty harness directory is a failure, not a pass" rule already in that comment, so I left it unguarded deliberately.

**The dependency row is fixed** — it claimed "only bash + jq" while the recipe hard-exits without `python3`, which since `test-scripts` joined `make check` is a contradiction a contributor hits at the pre-push gate. I put the `.cargo/config.toml` side effect (#455) next to it in the commands section rather than in the size-gate section, since it belongs to `test_setup_dev.sh` rather than to the size gate.

---

evansenter:
Round 26's three suggestions are all real, and all three are cost or diagnosability rather than verdict — so I've filed them as #456 rather than spinning CI a sixth time on this PR. Twenty-six rounds in, with the last blocking finding several rounds back, the marginal value of another cycle here is below the cost of one.

One correction worth making before it gets carried into #456's implementation. I measured the `basename` suggestion rather than taking the shape of it:

| form | 30 files |
|---|---|
| `basename` (current) | 57 ms |
| `${f##*/}` | 2 ms |
| predicate echoes the name, read via `$(...)` | 20 ms |

The first half is a 30× win. But the second half — having `is_example_binary` echo its derived name so both loops drop their own call — gives back about 90% of it, because command substitution forks a subshell per call. It's still better than `basename`, so it isn't a regression, but it isn't the version to take. The drift argument that motivated hoisting the filters doesn't extend to the name: the two loops derive it from the same `$f` with the same expansion, so there's nothing to drift. #456 says to take `${f##*/}` in all three places and leave the predicate a pure predicate.

On the glob guard — agreed the asymmetry is the point, since the loop below already answered the same question with a named message. Recorded that way.

---

Issue #430, opened 2026-08-16
Five response structs aren't `#[non_exhaustive]`, contradicting the documented convention


evansenter:
`docs/ENUM_WIRE_FORMATS.md:1236` states the convention:

> ### Structs and `#[non_exhaustive]`
>
> Response structs (e.g. `AutoFunctionResult`) also use `#[non_exhaustive]` so we can add fields without breaking user code.

Five response structs don't follow it:

| Struct | File |
|---|---|
| `Trigger` | `src/triggers.rs:274` |
| `TriggerExecution` | `src/triggers.rs:776` |
| `Environment` | `src/environments.rs:139` |
| `Agent` | `src/agents.rs:40` |
| `Webhook` | `src/webhooks.rs:312` |

None carries the attribute, so all five are constructible with struct literals outside the crate — and adding a field to any of them is a breaking change.

This just cost a break: #406 added an `extra` field to exactly these five, which would have been a non-event under the documented convention. The same will be true of the next field the API grows, which for these resources is a matter of when rather than if — that is the premise #406 was filed on.

The doc's own reasoning applies cleanly here:

> - Response types represent API responses, not user-constructed data
> - Mocking them in unit tests would give false confidence
> - We can add fields (like `executions` was added to `AutoFunctionResult`) without breaking changes

## Proposed work

- [ ] Add `#[non_exhaustive]` to the five structs
- [ ] Sweep for other response types missing it — treat the list above as found-so-far, not exhaustive
- [ ] Check in-crate construction sites still compile (`#[non_exhaustive]` only restricts *external* struct literals, so in-tree tests are unaffected)
- [ ] Note in `ENUM_WIRE_FORMATS.md` that the convention is enforced by review, or add a lint/test that catches a new response struct without it

## Cost

This is itself a breaking change for anyone constructing these types today — struct-literal construction stops compiling entirely, with no `..Default::default()` escape. That is the intended outcome per the doc's rationale, but it is worth landing in the same release as #406's field addition rather than as a second, separate break.

`FileSearchStore` / `FileSearchDocument` (added in #422) already carry `#[non_exhaustive]`, so the convention holds for new code — this is a backlog of five.

Issue #423, opened 2026-08-16
`with_cached_content()` references a resource the crate cannot create — no `cachedContents` CRUD


evansenter:
`InteractionBuilder::with_cached_content("cachedContents/xyz")` (`src/request_builder/mod.rs:1913`) sends the explicit-caching pointer, but the crate has no way to *produce* one. Same shape of gap as the File Search stores issue: we can reference the resource, not provision it.

The endpoint is live. Probed 2026-08-16 with a standard `GEMINI_API_KEY`:

- `GET /v1beta/cachedContents` → 200
- `POST /v1beta/cachedContents` `{"model": "models/gemini-3.7-flash", "contents": [...], "ttl": "300s"}` → 400 `Cached content is too small. total_token_count=2, min_total_token_count=1024`

The 400 is a content-size rejection, not a availability or auth one — the endpoint accepted the payload shape and validated it. A real create needs ≥1024 tokens of cached content.

## Proposed work

- [ ] `Client::create_cached_content()` / `get_` / `list_` / `update_` (TTL refresh) / `delete_`
- [ ] Typed `CachedContent` with a flattened Evergreen `extra`
- [ ] Note the payload uses the older `contents`/`parts` shape and `models/`-prefixed model names, not the Interactions API's `input`/steps model — worth documenting the seam explicitly so callers are not surprised
- [ ] Surface the ≥1024-token minimum in the docs; it is the first thing anyone hits
- [ ] Integration test gated on `GEMINI_API_KEY` (needs a ≥1024-token fixture)
- [ ] Cross-reference from `with_cached_content()`'s doc comment once creation exists

Lower priority than #422 — caching is an optimization, whereas file search is unusable without provisioning. But the two together are the same story: the crate models the Interactions API surface well and leaves the adjacent resource APIs to the caller.

evansenter:
Closing this, and the reason is more interesting than the issue.

I filed this to add CRUD so callers could produce a cache for `with_cached_content()` to point at. Before building it I probed whether the pointer works. **It doesn't** — the Interactions API rejects the field outright:

```
POST /v1beta/interactions  {"model": "...", "input": "...", "cached_content": "cachedContents/..."}
→ 400 Unknown parameter 'cached_content'
```

[...]

Issue #265, opened 2026-01-03
MCP Server: add builder auth helper, live integration test, and example (unblocked; core types landed)


evansenter:
## Summary

Complete the MCP (Model Context Protocol) Server tool implementation with authentication headers, proper content type handling, and builder convenience methods. This supersedes and expands upon #64.

## Problem / Motivation

We have a basic `Tool::McpServer { name, url }` skeleton, but it's incomplete:
1. **No authentication headers** - MCP servers often require auth
2. **Missing content types** - `mcp_server_tool_call` and `mcp_server_tool_result` not handled
3. ~~**No builder methods** - No convenient `.with_mcp_server()` method~~ ✅ Added in PR #295
4. **Limited config** - Can't configure timeouts, retries, etc.

## Context

- **Discovered during**: API feature completeness audit
- **Supersedes**: #64 (Add remote MCP support to Interactions API)
- **Relevant files**:
  - `genai-client/src/models/shared.rs` (Tool enum)
  - `genai-client/src/models/interactions/content.rs` (content types)
  - `src/request_builder/mod.rs` (builder methods)
- **API Reference**: [Interactions API Reference](https://ai.google.dev/static/api/interactions.md.txt)
  - Tool Types section:
    - "MCP Server - Model Context Protocol servers with authentication headers"
  - Content Types section:
    - `type: "mcp_server_tool_call"` - MCP tool invocation
    - `type: "mcp_server_tool_result"` - MCP tool response
- **⚠️ BLOCKED**: Remote MCP does not work with Gemini 3 models yet (Google says "coming soon")
  - Supported: gemini-2.5-pro, gemini-2.5-flash, gemini-2.5-flash-lite
  - NOT supported: gemini-3-flash-preview, gemini-3-pro-preview

### 1. Extend Tool::McpServer

```rust
// In genai-client/src/models/shared.rs
pub enum Tool {
    // ... existing variants ...
    
    /// Model Context Protocol (MCP) server
    McpServer {
        /// Server name (snake_case, no hyphens)
        name: String,
        /// Server URL (must be HTTPS)
        url: String,
        /// Optional authentication headers
        #[serde(skip_serializing_if = "Option::is_none")]
        headers: Option<HashMap<String, String>>,
    },
}
```

### 2. Add MCP Content Types

```rust
// In genai-client/src/models/interactions/content.rs
#[non_exhaustive]
pub enum InteractionContent {
    // ... existing variants ...
    
    /// MCP server tool invocation
[...]

PR #444, opened 2026-08-16
Add DECISIONS.md and CONTRIBUTING.md

### LINKED ISSUE #426: Add DECISIONS.md and relocate the design rationale currently scattered through CLAUDE.md and docs/
The repo has no `DECISIONS.md`, no ADR directory, and no `CONTRIBUTING.md`. The durable "why" behind design choices currently lives inside `CLAUDE.md` (292 lines) alongside operational instructions — build commands, test flags, the version-bump checklist — so rationale and procedure are interleaved with no separation.

## Why this is worth fixing

Two concrete symptoms, both already visible in the tree:

1. **`AgentBuilder`'s default model was swept to `"test-model"`.** The correction lives as a cautionary paragraph in CLAUDE.md ("None of this extends to non-test code: a fallback model id that reaches the wire must be a real one"). The rule is documented; the *decision* that produced it — why test code may use either form while non-test code may not — is inferable but never stated as such.
2. **#265's RFC points at `genai-client/src/models/shared.rs`**, a layout that no longer exists. The crate was restructured with no record of the move, so a 7-month-old issue silently references a phantom path.

## Scope: relocation, not authoring

The rationale is largely already written — this is mostly a move, which is what keeps the task bounded. Existing prose to draw from:

- `CLAUDE.md:42` — doctests run only in CI, and why
- `CLAUDE.md:266` — `-D warnings` on the docs.rs check is deliberately stricter than docs.rs itself
- `CLAUDE.md` — the model-constant table, `INLINE_VIDEO_MODEL` / `MINIMAL_THINKING_MODEL` live-verified justifications, and the `"test-model"` boundary
- `docs/ENUM_WIRE_FORMATS.md:1239` — why users cannot construct certain types directly
- `docs/ENUM_WIRE_FORMATS.md:1268` — a deliberate departure from faithful roundtripping
- `docs/TESTING.md:549` — test organization by feature rather than by pattern
- `docs/INTERACTIONS_API_GAP.md` — the "Spec-vs-implementation disagreements" section, which is a decision log in all but name
- `CLAUDE.md` — Evergreen soft-typing as the core philosophy, currently stated as a rule rather than a decision with alternatives considered

**Explicitly out of scope:** reconstructing ADRs from git archaeology across the 13 releases in CHANGELOG.md. That version is open-ended and would itself go stale — the exact failure mode this audit just found elsewhere. Seed the file with the decisions that already exist in prose, and add new ones going forward as they are made.

## Proposed work

- [ ] Create `DECISIONS.md` with a light template (context / decision / consequences / date; no heavy ADR ceremony)
- [ ] Relocate the rationale listed above, leaving pointers from CLAUDE.md so the operational docs stay navigable
- [ ] Leave CLAUDE.md as instructions-for-agents: commands, conventions, checklists
- [ ] Add `CONTRIBUTING.md` covering the `make check` gate, the live-verification protocol for wire changes, and when a change warrants a DECISIONS entry
- [ ] Note in DECISIONS.md that structural moves should be recorded, so the #265 situation is not repeated

evansenter:
Closes #426.

The durable "why" lived inside `CLAUDE.md` alongside operational instructions — build commands, test flags, the version-bump checklist — so rationale and procedure were interleaved with no separation, and some of it was scattered into `ENUM_WIRE_FORMATS.md` and `TESTING.md` where nobody would look for it.

## 11 decisions, relocated from existing prose

| | Decision |
|---|---|
| D-001 | Evergreen soft-typing |
| D-002 | Response structs are `#[non_exhaustive]` |
| D-003 | Antigravity alias enums break round-trip *deliberately* |
| D-004 | Verify from generated bindings first, prose last |
| D-005 | Spec-present but API-rejected: remove vs keep for parity |
| D-006 | Model ids live in constants only |
| D-007 | Breaking changes over compatibility shims |
| D-008 | Tests organized by feature, not pattern |
| D-009 | Doctests in CI only; docs.rs check deliberately stricter |
| D-010 | A structural test is not evidence a feature works |
| D-011 | Record structural moves |

Scope was deliberately limited to relocation. Reconstructing decisions from git archaeology across 13 releases is open-ended and would go stale — the exact failure mode this repo just spent a day fixing elsewhere.

Two go beyond relocation, and they're the two that most needed writing down:

- **D-010** — `cached_content` was marked ✅ with a passing serialization test and was rejected by the API for its entire shipped life (#423). The general lesson outlives the specific bug.
- **D-011** — the direct answer to this issue's own #265 example, citing it as the cost of not recording moves.

**D-003 is filed as a deliberate departure from D-001**, not an exception. A principle with an unexplained hole invites the next person to widen it.

**D-002 carries a "known gap"** pointing at #430 — five response structs don't follow the convention it states. Better to record the rule with its violations named than to state a rule the code contradicts.

## CONTRIBUTING.md

Setup, the gate, the verify-from-bindings-first ordering, a checklist for adding wire types, and when a DECISIONS entry is warranted. Documents the mold linker trap from #428 explicitly, since the error message names the wrong tool and costs a newcomer real time.

`CLAUDE.md` keeps the operational rules and gains a pointer to the reasoning; `README.md` gains a Project section. All D-0NN cross-references resolve, and doctests pass (both files are included via `doc_comment!`).

---


evansenter:
Both round-9 items fixed in `a448d26`, and you were right that the linker one was the merge-blocker of the two — it is the first section a new contributor reads, and it was sending them to work around a failure that no longer happens.

Verified each claim against the merge head rather than taking the diagnosis:

```
make check target:      check: fmt clippy test test-scripts
setup-dev.sh exists:    yes
config.toml tracked:    0        (.gitignore:13; only .cargo/config.toml.example is tracked)
```

The Setup section now points at `./scripts/setup-dev.sh` and matches `CLAUDE.md`. I kept the `cannot find 'ld'` failure text, since an old checkout carrying the generated config still hits it — but the remedy is now "delete `.cargo/config.toml`" rather than the `CARGO_TARGET_..._RUSTFLAGS=""` override, which on a current clone would be clearing rustflags that are not set. The `(Tracked as #428 …)` parenthetical is gone.

On the gate: took your point about the audience. Adding `test-scripts` to the command line alone would have been half a fix, since the consequence is that `make check` now has two prerequisites Setup never listed and that target hard-fails rather than skipping — so `jq` and `python3` are now listed under Setup with that reason stated.

---

Issue #433, opened 2026-08-16
`Step::McpServerToolCall`/`McpServerToolResult` appear unreachable — the API emits generic `tool_call` steps


evansenter:
Found while writing the live MCP test for #265. The crate models `mcp_server_tool_call` and `mcp_server_tool_result` step types (added per #265's checklist, from the spec). The API doesn't emit them.

## Observed

Live 2026-08-16, `gemini-3.7-flash`, a real MCP server (`mcp.deepwiki.com`), successful call:

```
type= tool_call  keys= ['id', 'signature', 'type']
type= tool_call  keys= ['id', 'signature', 'type']
```

Through the crate, the same interaction:

```
MCP status: Completed
Steps: ["tool_call", "model_output", "tool_call", "model_output", "thought", "model_output"]
Tool-use tokens: 1168
```

`tool_call` is not a modeled `Step` variant, so those steps land in `Step::Unknown { step_type: "tool_call", .. }`. Evergreen degrading exactly as designed — nothing breaks, the interaction completes, the answer is correct.

## What it costs

- `Step::McpServerToolCall` and `Step::McpServerToolResult` look like dead code. Their fields (`server_name`, `arguments`, `call_id`) have never been observed on the wire.
- `step_summary().mcp_server_tool_call_count` and `.mcp_server_tool_result_count` read **0** on a successful MCP call, which is worse than absent — a caller checking them concludes MCP didn't run.
- The generic `tool_call` step carries only `{id, signature, type}` — no server name, no tool name, no arguments. So there is currently no way to tell *which* MCP server or tool was invoked from the response.

Same shape as #429 (`file_search_result` steps carrying no chunks): a spec-derived type modeling richer data than the endpoint returns.

## Options

1. **Model `tool_call` as a known step type.** Makes the common case first-class instead of `Unknown`. Small and clearly correct regardless of what happens to the MCP-specific variants.
2. **Keep the MCP variants, document them as spec-only.** Same treatment as `Tool::Retrieval` — present for parity, noted as unobserved. Cheap, honest.
3. **Fix the step-summary counters** so they either count `tool_call` steps or stop claiming zero. The current behavior is the actively misleading part.
4. **Remove the MCP variants.** Follows the `cached_content` precedent (D-005), but premature here: unlike `cached_content`, nothing *rejects* them — they simply haven't been seen, and the API may start emitting them.

[...]

Issue #306, opened 2026-01-08
Add tests for Computer Use tool when API becomes available


evansenter:
## Summary

The `with_computer_use()` and `with_computer_use_excluding()` builder methods exist but have no integration tests.

## Current State

- `with_computer_use()` - implemented in `src/request_builder/mod.rs:1443`
- `with_computer_use_excluding()` - implemented in `src/request_builder/mod.rs:1493`
- Wire format marked as **UNVERIFIED** in `docs/ENUM_WIRE_FORMATS.md`
- No tests exist

## Why Blocked

Computer Use has limited API access and the wire format is unverified. We need:
1. API access to test against
2. Verification of the wire format with `LOUD_WIRE=1`

## When to Unblock

- When Google makes Computer Use generally available
- When we can verify the wire format

## Action Items

- [ ] Verify wire format with `LOUD_WIRE=1` when available
- [ ] Add unit tests for builder methods
- [ ] Add integration test demonstrating usage
- [ ] Update `docs/ENUM_WIRE_FORMATS.md` with verified format

evansenter:
**Unblocked** — the Computer Use tool is live on a standard `GEMINI_API_KEY`.

Verified 2026-08-16 against `gemini-3.7-flash`:

```
POST /v1beta/interactions
{"model":"gemini-3.7-flash","input":"go to example.com",
 "tools":[{"type":"computer_use","environment":"browser"}]}
```

→ `200`, `status: "requires_action"`, with computer-use steps and a thought signature returned. That is the tool functioning normally — `requires_action` is the expected state, since the caller is meant to execute the action and return the result.

Removing the `blocked` label; this is now ordinary test work. `examples/computer_use.rs` already exists, so the example scaffolding is in place.

Found during a stale-issue / feature audit.

---
_Generated by [Claude Code](https://claude.ai/code)_

evansenter:
Test added on branch `claude/issue-265-306-tool-tests`, uncommitted to remote pending review. **Passing live.**

```
Computer use status: RequiresAction, steps: ["thought", "function_call"]
test computer_use::test_computer_use_reaches_requires_action ... ok
```

The assertion is `requires_action`, deliberately **not** `completed`. The model emits an action for the caller to execute and waits — so asserting on the usual success status would invert the test, passing exactly when the tool was ignored. That's the kind of detail that only surfaces by running it.

[...]

PR #445, opened 2026-08-16
Add live integration tests for MCP and Computer Use

evansenter:
Addresses #306 and the test scope of #265.

Both tools were recorded as blocked — #265 ("Remote MCP does not work with Gemini 3 models yet") and #306 ("when API becomes available") — and both work today. A live test is the only thing that would have noticed either becoming available, which is why they sat blocked for months.

## MCP

Asserts the interaction completes and that the server was **actually called**. Tool-use token count is the load-bearing assertion rather than the answer text, which would pass on a model that ignored the tool and guessed from its own knowledge.

Verified live against `mcp.deepwiki.com` on `gemini-3.7-flash`:

```
MCP status: Completed
Steps: ["tool_call", "model_output", "tool_call", "model_output", "thought", "model_output"]
Tool-use tokens: 1168
```

## Computer Use

Asserts `requires_action`, **not** `completed`. The model emits an action for the caller to execute and waits; asserting on the usual success status would invert the test, passing exactly when the tool was ignored.

```
Computer use status: RequiresAction, steps: ["thought", "function_call"]
```

## Failure discipline

Both distinguish a real regression from an environmental one and skip only on the latter — MCP on the third-party server being unreachable, computer use on the tool not being allowlisted for the key. An API rejection of the tool itself panics, since that is what these exist to catch.

Wire-shape unit tests alongside each, needing no key. The computer-use one pins `excluded_predefined_functions` in snake_case, which was a spec-vs-implementation disagreement once already.

## Finding: MCP step variants look unreachable

The API emits generic `tool_call` steps for MCP, not the `mcp_server_tool_call` / `mcp_server_tool_result` types the crate models from the spec. On the wire they carry only `{id, signature, type}`, so they deserialize into `Step::Unknown` and `step_summary().mcp_server_tool_call_count` reads **0** on a successful call.

Evergreen degrading as designed, but two modeled variants have never been observed and a counter reporting zero is worse than one that's absent. Filed as #433. The assertion here is written on tool-use tokens precisely so it survives whichever way the step typing lands.

---


evansenter:
Round 11 addressed in 6edf13e. Two taken as written, one taken in part.

**1. `release.yml` counts both markers.** Confirmed the premise before changing it: that step runs `cargo nextest run --workspace --run-ignored all`, so `test_mcp_server_tool_round_trip` does execute there and could emit the marker uncounted. Switched to the same `-cE` alternation and widened both the `::error::` and `::warning::` text off "semantic validation(s)" to "assertion(s)", since the count is no longer only semantic ones. The threshold comment now reads "~48 semantic validations plus the two tool-evidence sites" — at two possible extra sources, 10 still sits where it was placed.

**2. `docs/TESTING.md` names both**, as a small table, with the distinction spelled out — validator call failed vs. interaction succeeded without evidence the tool ran. `rust.yml`'s `--success-output=final` comment now names both markers too.

**3. The two prose skips — split.** Your counter-argument is the deciding one, but it separates them rather than covering both:

- **MCP unreachable: marked.** This branch is the catch-all `else` of the error arm — it fires on anything `about_mcp && rejected` did not recognise. So it genuinely cannot distinguish a dead deepwiki from a rejection phrased in words that guard misses, which is the same ambiguity the Ok-path marker was added for. A permanently unreachable server now shows up as one marker in the `tools` group: a warning, not a failure, which is the right weight for a third-party outage.
- **Computer use: left plain, with a comment saying so.** Both halves of that guard must match, so it fires only on an error naming computer use *and* saying it is unavailable; everything else in the arm panics. That is a precisely-scoped environment property, closer to the `GEMINI_API_KEY`-unset skip than to a vacuous assertion — there is no regression a green run could be hiding. Marking it would annotate every run on an un-allowlisted key forever, which is noise. The docs paragraph records the boundary in general terms so the next skip site has a rule to apply.

Verified: both workflows parse, the alternation counts a mixed log correctly (3 for 2× tool-evidence + 1× semantic), and `cargo fmt --check` and `clippy --workspace --all-targets --all-features -D warnings` are clean.

---

evansenter:
Round 12 addressed in 6f0bc58. All three taken, one of them the opposite way round from the suggested alternative.

**1. `rust.yml` keeps `-gt 3`, and now says why.** The tightening is intended, so the comment states it rather than the threshold moving. A permanently unreachable MCP server does consume one of `tools_and_config_tests`'s three slots — but a run that is *both* missing live tool evidence and burning validator transients is precisely the run whose green is least worth trusting, so spending them from one pool is the behaviour I want. `-gt 4` would also have loosened every other matrix group, none of which emits the new marker. The comment now contrasts this with `release.yml`, which re-derived its number because it counts one whole-workspace run against a far larger denominator, where one extra marker really is noise.

**2. The `Err`-arm marker prose.** Correct catch — the text asserted a diagnosis the branch's own comment says it cannot make. Now:

```
LIVE_TOOL_EVIDENCE_SKIPPED: MCP call failed for an unrecognised reason (server down, or a rejection this guard did not match): {e:?}
```

Compiled and ran the exact `println!` to confirm the `\` continuation still collapses to one output line: `lines: 1  marker count: 1` under the same `grep -cE`.

**3. "the two tool-evidence sites" → "at most one".** Right, and for the reason given — both sites are in `test_mcp_server_tool_round_trip` and are mutually exclusive, the `Err` arm returning before the Ok path is reachable. The comment now spells that out, which does make it self-evident why 10 did not need to move.

`cargo fmt --check` and `clippy --workspace --all-targets --all-features -D warnings` clean; both workflows parse.

---

evansenter:
Round 13 addressed in 229a784. All three taken.

**1. `"not available"` — kept, and the comment now says the trade was made knowingly.** The exposure is real and the shape is the same one `"invalid"` was excluded for. Keeping it anyway, for the reason you named: the false-positive phrasing is speculative, while the miss it prevents is not — an unrecognised rejection reported as a skip is a real regression going quiet, and this panic is loud and diagnosable when it misfires. Rounds 8 and 9 pushed the guard toward catching rejections; this stays consistent with that rather than quietly walking it back.

**2. The three doc sites now point at `response.tool_use_tokens()`.** Right that the argument applies more strongly in the docs than in the test — the docs are where a caller learns which surface to use, and the accessor is where the `Some(0)`-versus-`None` rustdoc lives. `docs/BUILT_IN_TOOLS.md`, `docs/ENUM_WIRE_FORMATS.md` and the `mcp_server_tool_call_count` rustdoc all updated; the rustdoc one is an intra-doc link (`[`InteractionResponse::tool_use_tokens`]`) so it is now checked by the doc build rather than being prose that can rot.

**3. Annotation wording.** Both workflows, all four lines:

```
$skips unverified assertion(s) or tolerated failure(s)
```

"Passed vacuously" was accurate while the counter covered only skipped assertions. The `Err`-arm marker is a tolerated failure where no assertion was reached at all, so the old phrasing had become wrong for one of the two classes it now counts.

`cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings`, and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` all clean; both workflows parse.

---

evansenter:
Round 14 addressed in 5df2d46.

Fair catch, and it's a consequence of round 13 rather than something that was wrong before it: once the three referring sites stopped naming the raw field and started pointing here, this rustdoc became the destination — and it carried only the half that invites the mistake. Added the same clause the three sites already agree on:

> It is a single aggregate across all tools, though, so it identifies *which* tool only when one is declared.

Which is exactly the condition the MCP test relies on and had recorded nowhere in the API docs a hover reaches.

`cargo fmt --check` and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean.

---

evansenter:
Round 15 addressed in d95bd01. All three taken.

**1. The `UsageMetadata::total_tool_use_tokens` field doc.** Correct — same shape as round 14, one layer down, and this was genuinely the last hover destination still making the unbounded claim on its own. It now carries the qualifier and points at `InteractionResponse::tool_use_tokens` as the caller-facing accessor with the fuller story, so the field doc and the accessor doc no longer disagree about how much you can conclude.

**2. `docs/TESTING.md` stated the wrong rule — this is the one worth having caught.** I wrote "stability" and that is not the line the code draws: a permanently unreachable `mcp.deepwiki.com` is every bit as stable a property of the environment as an un-allowlisted key, and it *is* marked. The rule actually applied is distinguishability, exactly as you say. Rewritten:

> The rule for whether a skip gets a marker is **distinguishability, not stability**. A skip whose guard is narrow enough that it cannot be a disguised regression is left unmarked […] A skip reached from a catch-all branch is marked, however stable its cause may turn out to be.

with the MCP `Err` arm named as the worked example. The old phrasing would have told the next skip site to leave a catch-all unmarked, which is the opposite of what this PR decided.

**3. Re-wrapped the ~130-column blockquote line.**

`cargo fmt --check` and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean.

---

evansenter:
Round 16 addressed in 7d3f3a1. All three taken.

**1. The computer-use guard was overstated — you're right, and the test's own comment is the evidence.** "Nothing else can reach it" is false: a model-level "computer use is not supported for this model" satisfies `about_computer_use` via the prose register and `unavailable` via `"not supported"`, so exactly the #306-reverting regression this test exists for would land in the unmarked branch. What rounds 8–9 actually bought is exclusion of *unrelated* 4xx wording, which is the narrower claim. Reworded to that, plus an explicit paragraph recording the gap:

> That scoping is deliberately not airtight, and the test says so: a model-level "computer use is not supported for this model" satisfies both halves too, so a regression of that kind would land in the unmarked branch. Leaving it unmarked is a recorded call, not an oversight — the alternative annotates every run on an un-allowlisted key indefinitely.

Agreed this is docs accuracy, not a request to change the guard, and I've not relitigated the branch staying unmarked. The point that mattered is that the prose is now where the next skip site learns the rule, and as written it taught something stronger than the code delivers.

**2. `rust.yml:356` marker comment.** Right — it predated the `Err`-arm emitter and still said "completed without evidence". Now carries the same two-case wording `docs/TESTING.md` and the round-13 annotation rewording already use, so all three sites agree.

**3. Re-wrapped the orphan line** the previous rewrap left mid-paragraph.

`rust.yml` parses.

---

Issue #406, opened 2026-08-09
Evergreen `extra` passthrough on response-side resource shapes (Trigger, TriggerExecution, Environment, Agent, Webhook)


evansenter:
Follow-up from a #405 review suggestion (`src/triggers.rs:274`).

#405 added a flattened Evergreen `extra` map to the request bodies (`CreateEnvironmentRequest`, `TriggerCreateParams`, `TriggerUpdate`), but the **response** shapes still drop unmodeled fields on deserialize, so a deserialize-then-serialize roundtrip loses them:

- `Trigger` / `TriggerExecution` — the sharpest case: the resource has no live-verified response shape while creation is agent-gated, so a field the API actually returns today is both invisible and unrecoverable to a caller
- `Environment`
- `Agent` / `Webhook` (pre-existing, same class)

`Content` and `RemoteEnvironment` already carry a flattened `extra` for exactly this reason. Adding it to one resource alone trades one inconsistency for another, so the right scope is all five response shapes in one pass:

- `#[serde(flatten)] pub extra: serde_json::Map<String, serde_json::Value>` on each
- roundtrip tests (unknown field survives deserialize → serialize)
- note: `PartialEq` derives already include the map; empty maps compare equal, so existing tests should be unaffected
- document the colliding-key precedence (flattened map wins on serialize), matching the request-side docs added in #405

Filed as a follow-up rather than in #405 per the reviewer's own scope note.

evansenter:
Implemented on branch `claude/issue-406-response-extra`, uncommitted to remote pending review. All five shapes done in one pass as the issue specified.

Roundtrip tests per shape cover both directions — an unknown field survives deserialize → serialize, and an empty map adds no key on serialize. Two extras beyond the checklist:

- The `Agent` test exercises the flatten through the **list envelope**, not just a directly-deserialized agent, since that is the path a real caller hits.
- A `Trigger` test pins that `PartialEq` over the added map still holds for identical wire, confirming the issue's note that existing equality-based tests are unaffected.

Colliding-key precedence is documented per field, matching the request-side wording from #405.

One correction to the issue's expectations: the `Webhook` fixtures need `uri` and `subscribed_events` included. Both are non-optional with no `skip_serializing_if`, so they always serialize and a minimal `{"id": "wh_1"}` fixture cannot roundtrip exactly. My first draft asserted otherwise and failed — pre-existing struct behavior, not a regression, but worth knowing if anyone writes similar fixtures.

[...]

PR #446, opened 2026-08-16
Make response structs `#[non_exhaustive]`, and enforce it with a test

### LINKED ISSUE #430: Five response structs aren't `#[non_exhaustive]`, contradicting the documented convention

evansenter:
Closes #430.

`docs/ENUM_WIRE_FORMATS.md` has stated the convention for a while, but 31 deserializable response types had drifted from it — including the five the issue names, which #406 had just turned into a breaking change by adding `extra` to them. Under the convention that addition would have been a non-event.

## Scope

Wider than the issue's five, narrower than "everything that derives `Deserialize`". The issue asked for a sweep and said to treat its list as found-so-far; the sweep turned up 119 candidates, most of which should *not* get the attribute.

**Added** — types the API returns:

| Group | Types |
|---|---|
| The five named | `Trigger`, `TriggerExecution`, `Environment`, `Agent`, `Webhook` |
| List wrappers | `AgentListResponse`, `EnvironmentListResponse`, `TriggerListResponse`, `TriggerExecutionListResponse`, `WebhookListResponse`, `ListFilesResponse` |
| Usage / response | `InteractionResponse`, `UsageMetadata`, `ModalityTokens`, `GroundingToolCount`, `OwnedFunctionCallInfo` |
| Result items | `ReviewSnippet`, `GoogleSearchResultItem`, `Place`, `GoogleMapsResultItem`, `UrlContextResultItem`, `FileSearchResultItem` |
| Files API | `FileMetadata`, `FileError`, `VideoMetadata`, `FileUploadResponse` |
| Misc | `SigningSecret`, `RotateSigningSecretResponse`, `StepError`, `StreamMetadata`, `StreamError` |

**Not added** — request-side types (`GenerationConfig`, `SpeechConfig`, `FunctionDeclaration`, the tool configs, the create/update bodies, …). A caller builds those. Closing them removes struct-literal construction — with no `..Default::default()` escape for a downstream crate — while buying the crate nothing, since it does not gain the freedom to add required fields to something the user assembles.

**Not touched** — `src/antigravity/protocol.rs`. That is the localharness JSON-RPC protocol, a separate wire format whose types the crate *sends* as well as receives, so this response/request split doesn't apply to it.

## Enforcement

The issue offered "note that the convention is enforced by review, or add a lint/test". Review is a poor detector for an attribute being *absent* — that's exactly how these five drifted — and the cost is invisible at the point it happens, surfacing a release later when a field the API grew turns out to be breaking.

So `tests/non_exhaustive_responses.rs` scans `src/` and fails on any deserializable public struct missing the attribute that isn't listed in its `REQUEST_SIDE` exemption list. Same shape as `tests/model_literals.rs`, including living in its own test binary so it doesn't re-run once per integration crate. Exempting a type requires adding it to a list that explains why.

## Fallout

Both cases are the change working as intended, not workarounds:

- Three strategies in `tests/proptest_roundtrip_tests.rs` built responses with struct literals. They now go through JSON — which is what that file's own module doc already prescribes for `AutoFunctionResult`. Nulls are dropped so absent stays absent rather than becoming an explicit null the roundtrip would then disagree on. All 24 proptests pass.
- One doctest hand-built a `FileSearchResultItem`. It now uses the `new()` that already existed, matching how the sibling item types are constructed in the same module.

## Verification

```
cargo test --lib                                    858 passed
cargo test --doc                                    all doctests ran
cargo test --test non_exhaustive_responses            1 passed
cargo test --test proptest_roundtrip_tests           24 passed
```

I also confirmed the new test actually discriminates: before the annotations it named all the offenders with file:line, and it caught its own name-parsing bug on `FunctionCallInfo<'a>` rather than passing over it.

## Note on release sequencing

The issue observes this is itself a breaking change and is best landed in the same release as #406's field addition rather than as a second separate break. That still holds — #436 (#406) is open, and both are in `[Unreleased]`.

---


evansenter:
All eight round-1 suggestions applied in 990fa2f.

**The `Default`-then-assign point was the important one, and I had it wrong.** I claimed in both the doc and the PR body that downstream crates can't construct these. `#[non_exhaustive]` removes struct-literal and `..Default::default()` syntax — it does not remove `T::default()` followed by field assignment. I verified by building an external crate that does exactly that against `UsageMetadata` and `InteractionResponse`; it compiles.

Consequences, all taken:

- `docs/ENUM_WIRE_FORMATS.md` now shows both forms and names the four types that genuinely need a JSON fixture, instead of sending every reader to one.
- Two of the three proptest strategies drop the JSON detour. You're right that it was also *lossier* — it pushed every generated value through `Serialize`+`Deserialize` before the roundtrip assertion saw it, which is precisely what the assertion is supposed to be testing.
- `ModalityTokens::new()` added, since it has neither `Default` nor a constructor.

**Guard hardening — all four blind spots.** The manual-`Deserialize` one was not hypothetical: it found two real cases the derive-only scan missed.

- `InteractionStreamEvent` — a genuine response type, now annotated.
- `RemoteEnvironment` — request-side, exempted with a note.

Also done: scan root anchored to `CARGO_MANIFEST_DIR` and asserted to exist (plus an empty-file-list failure), `path:Name`-scoped exemptions, `pub struct` matched after trimming, and paren-depth tracking in the walk-back so multi-line `#[derive(` blocks are read.

And a second test that runs the scan over a fixture covering all seven cases — plain-missing, manual impl, multi-line derive, indented-in-mod, already-annotated, serialize-only, and exempt — plus an assertion that the same name at a different path is *not* exempt. Your point that the validation lived in a PR description where nothing keeps it true was exactly right; when I fixed the walk-back afterward, that test is what confirmed it.

**CHANGELOG.** You caught two things here. There was no entry, and my PR body's claim that "both are in `[Unreleased]`" was wrong — 0.10.0 is the top entry and there was no `[Unreleased]` section at all, so my edit had silently no-oped. Now opens one, with the default-then-assign migration spelled out. (I checked #447 and #448 for the same silent no-op; both had it, both fixed.)

---

**On the CI failure.** `Integration Tests (multimodal)` is red on infrastructure, not on this diff:

```
info: downloading .../cargo-nextest-0.9.143-x86_64-unknown-linux-gnu.tar.gz
curl: (35) Recv failure: Connection reset by peer     ×11
##[error]Process completed with exit code 35
```

The job died in `taiki-e/install-action` before any test body ran, and the four sibling integration jobs on the same run downloaded the same artifact fine. `Check`, `Clippy`, `Format`, `Documentation`, `MSRV`, `Code Coverage` and `Integration Tests (files-and-wire | functions)` are all green on this SHA.

I tried to re-run the failed job and got `403 Resource not accessible by integration`, so that needs a maintainer click — or it'll clear on the next push. Flagging rather than silently leaving it red.

---

evansenter:
All four round-2 suggestions applied in f1bb82a.

**Doc prose in the attribute block.** Taken, and this was the strongest of the four — the guard could be satisfied by *text* rather than by an attribute, which is precisely the silent pass the file argues against. Both checks and the paren accounting now see attribute lines only.

Two things worth reporting from doing it:

1. Your unmatched-paren analysis was right about the mechanism, and it's worse than a false negative — the walk-back doesn't just misjudge that struct, it runs off into earlier source that almost certainly contains `non_exhaustive` somewhere.
2. **The fixture test caught a regression in my own fix.** Excluding doc lines also excluded the `)]` that *ends* a multi-line derive, which the walk-back meets first — so `MultiLineDerive` stopped being detected and the test went red. That's exactly the "keeps a later refactor from turning the guard into a no-op that still reports green" case it was added for, one round after it was added.

**Stale `REQUEST_SIDE` entries.** Taken. Your framing of the two directions is what makes it worth enforcing: a *moved* type surfaces on its own (it gets scanned unexempted), but a *deleted* one leaves a key that pre-authorizes whatever future type lands on that `path:Name`. Since each entry carries a written justification, a stale one is a justification for something that doesn't exist. Every key is now asserted to have matched during the scan.

**The JSON-fixture list.** Short by four and wrong by one, as you said — `FileError`, `VideoMetadata`, `ListFilesResponse` and `InteractionStreamEvent` added; `OwnedFunctionCallInfo` removed, since `FunctionCallInfo::to_owned()` is public and produces one. I also took your point about `InteractionStreamEvent` on its own terms and gave it the same treatment `ModalityTokens` got.

**`ModalityTokens::new()` unit test.** Taken — CLAUDE.md is explicit and I missed it. Covers both `Into<String>` forms, since the signature is what a caller actually sees.

**On the PR body.** You're right that the Fallout section still describes the JSON-with-nulls-dropped shape from round 1. I'll leave the body as the historical record of what was proposed and let this thread carry the correction, rather than quietly rewriting it to match — the round-1 approach being wrong is part of the story.

**CI.** `Integration Tests (multimodal)` is still the `install-action@cargo-nextest` download failure (`curl: (35)`, ×11, before any test ran), not this diff — and thank you for independently confirming that read. I can't re-run it (`403 Resource not accessible by integration`); the push above should re-trigger.

---

evansenter:
All four round-3 items applied in ff2c52c.

**The inert `FunctionCallInfo` exemption — and the rule that couldn't see it.** This was the best finding of the round, because the flaw was in the check I'd just added. You're right that staleness meant "no longer names a type" when it needed to mean "no longer excuses anything". Recording a match only when the exemption actually suppresses an offender fixed the rule, and it immediately reported:

```
These REQUEST_SIDE exemptions no longer match any type:
  src/response.rs:FunctionCallInfo
```

exactly as you predicted. Removed. `FunctionCallInfo` is `Serialize`-only so the scan never would have reported it — the entry existed purely to pre-authorize a future `Deserialize`, silently.

**Leading `)` at depth 0.** Taken, and your two-directional framing is what makes it worth tightening rather than leaving: swallowed `non_exhaustive` reads as annotated (silent pass), swallowed `Deserialize` reads as a response type (false positive). Now matches `)]` exactly. I appreciate you checking all 177 declarations to establish it isn't reachable today rather than just flagging it as theoretically possible.

**Stale check had no fixture coverage.** Right, and it was the one branch that didn't — the fixture went through `offenders_in`, which discards the set. It now asserts the exemption set holds exactly `src/request.rs:GenerationConfig` on one path and is empty on another, which also pins path-scoping from the direction the existing assertions didn't reach.

**`InteractionStreamEvent` `new()`.** Factual error on my part — I confused it with `StreamEvent`, which does have one, exactly as you guessed. Moved to the JSON-fixture list and the sentence corrected.

Also noting your check that all 29 keys resolve, including the two awkward ones — `AgentConfig` as a tuple struct and `FunctionCallInfo` as generic. Those are precisely the two the name-splitting could have got wrong, and confirming the new assertion doesn't fire spuriously on its own first run is the check I'd have wanted.

```
cargo test --lib                       859 passed
cargo test --test non_exhaustive_responses   2 passed
```

---

evansenter:
Both round-5 findings fixed in 30f7cb8. The Important one was a real bug in the previous round's change, and your trace of it was exact.

**The counter never unwound.** Seeded to `1` on the *attribute* line, so the gated item's balanced braces returned it to `1`, not `0` — and since the rest of the file is brace-balanced it stayed there to EOF. Your two concrete cases were right: `src/serde_util.rs` dropped out from line 45, `src/lib.rs` from line 60, both of which the round-4 scanner read in full.

What makes this worth calling out beyond the fix: the guard covered *less than the round before* while still reporting green. That is exactly the silent-inertness failure this file exists to catch, occurring in the file that exists to catch it.

The fix needs two states, because a gated item may or may not have braces — `#[cfg(test)] mod tests { .. }` and `#[cfg(test)] resource_markers! { .. }` are brace-delimited, `#[cfg(test)] pub(crate) mod test_subscriber;` ends at a semicolon. One counter can't express both, which is why the single-counter version was doomed rather than just off-by-one.

**And the inner-attribute check matched nothing.** You're right, and it's the more embarrassing half: the comment named the five `*_tests.rs` files while the code checked for `#[cfg(test)]`, which none of them has — the gating is on the `mod NAME;` lines in `lib.rs`. Now derived from `lib.rs` as you suggested, so a sixth test module is picked up automatically, with an assertion that fails loudly if that declaration pattern ever changes rather than silently classifying nothing.

**Fixture, and proof it discriminates.** Added a `Deserialize` struct below the cfg-test block and another below a semicolon-gated item — the positions you identified as the only ones that can catch this, since the old fixture put the cfg-test block last. Then I reverted the fix to confirm:

```
scanning must resume after a cfg(test) block — a counter that never unwinds
skips the rest of the file: ["src/request.rs:7: Bare", ..., "src/request.rs:30: ImplInAnotherFile"]
test result: FAILED

# with the fix restored
test result: ok. 2 passed
```

```
cargo test --lib                             859 passed
cargo test --test non_exhaustive_responses     2 passed
```

---

evansenter:
Round 9's three suggestions applied in `5522f22`. All three were right, and the second one was worse than described.

**Module-root keying.** `cfg_test_modules` now takes the file path and derives the root as you specified — `dir` for a `mod`/`lib` stem, `dir/stem` otherwise. Fixture extended with a leaf case: `mod helpers;` in `src/http.rs` keys to `src/http/helpers`, and `src/helpers.rs` is explicitly asserted *not* skipped.

**Inline blocks.** Declarations must now end in `;` and parse as a bare identifier. Worth noting the four reserved names were not merely useless — the round-8 commit message counted them as evidence the crate-wide walk was working, so they were actively misleading about coverage. `src/http/proptest_tests` is the one that would have bitten: `src/proptest_tests.rs` already exists, so a sibling under `src/http/` is a name the crate demonstrably reaches for.

**The expected list.** Added `src/request_builder/tests`. Your read is exactly right — I confirmed the old list was inert against everything round 8 built by reverting the collection to a `lib.rs`-only read, and the loop stayed green.

That revert is one of three mutations I ran rather than reasoning about, since the whole argument of this file is that a guard nobody broke on purpose is a guard nobody has tested:

```
CAUGHT  - module root reverts to the declaring directory   (skip-scope fixture)
CAUGHT  - inline mod blocks collected again                (skip-scope fixture)
CAUGHT  - collection reverts to lib.rs only                (main test)
```

Each fails an assertion rather than a compile error — I re-ran the middle one after the first attempt produced a type error instead, which would have been a false positive.

`fmt`, `clippy --all-targets --all-features -D warnings`, and the full non-integration suite (859 lib + all integration binaries) green locally.

---

PR #447, opened 2026-08-16
Make `#[tool]` need nothing but `genai-rs` and `genai-rs-macros`

### LINKED ISSUE #402: Macro: emit fully-qualified paths so #[tool] needs no consumer-side imports

evansenter:
Closes #402.

The macro imposed two requirements on consumers, neither of which is visible from the `#[tool]` attribute itself:

1. **Two direct dependencies.** The generated code named `::async_trait` and `::serde_json`. A leading `::` resolves in the *consumer's* dependency graph, so every downstream crate had to add both to its own `Cargo.toml` just to use the macro.
2. **A trait import at every expansion site.** `declaration()` was called in method position, which needs `use genai_rs::CallableFunction;` in scope wherever `#[tool]` is applied.

## The fix

- A `#[doc(hidden)] pub mod __private` re-exporting `async_trait` and `serde_json`, with the macro emitting paths through it. These now resolve in *this* crate's graph.
- `declaration` called fully-qualified: `<Callable as CallableFunction>::declaration(&…)`.

## Verification

I built an actual fresh consumer crate with exactly two dependencies and no trait import, rather than reasoning about what would happen.

**Before** — three `E0433`s and an `E0195`, all pointing at the `#[tool]` line:

```
error[E0433]: failed to resolve: could not find `async_trait` in the list of imported crates
error[E0433]: failed to resolve: could not find `serde_json` in the list of imported crates
error[E0433]: failed to resolve: could not find `serde_json` in the list of imported crates
error[E0195]: lifetime parameters or bounds on method `call` do not match the trait declaration
```

**After** — clean build.

```
cargo test --lib            858 passed
cargo test --doc            all doctests ran
cargo test --test macro_tests   9 passed
cargo test --test ui_tests      4 fixtures, all ok
```

## Regression guard

`tests/ui/pass_no_consumer_imports.rs` — a `trybuild` pass fixture that uses `#[tool]` with no import beyond the macro, and exercises both generated entry points.

Worth being explicit about why a `pass` fixture and not another `compile_fail`: the existing suite could never have caught this class. A macro that demands *more* of its caller still fails every case it is supposed to fail, so `compile_fail` stays green through exactly this regression.

The fixture covers the trait-import half directly. It cannot cover the missing-dependency half, since trybuild builds against genai-rs's own dev-dependency graph where both crates are present — the `consumer-crate` workflow in #442 is what covers that, and its `minimal-requirements` job was filed tracking this issue specifically. Once both land, that job can flip from `continue-on-error` to a gate.

## Doc sweep — partial by intent

The README's "required together" dependency grouping and the import note are gone, and the two snippets whose `CallableFunction` import was genuinely unused (`SECURITY.md`, `examples/real_world/data_analysis/README.md`) lose it.

Snippets that call `.declaration()` in method position **keep** their import. That is a real trait-method call and still needs the trait in scope — it is not a macro requirement. Rewriting them to use the generated `<fn>_declaration()` free function instead would be a readability improvement, but it is a separate change from removing the requirement, and doing it blind across ~15 snippets risks breaking doctests for no gain here.

---


evansenter:
All four applied in 3fb78a9.

**Version coupling.** The sharpest of the four, because it's a hazard this PR *creates* — you're right that the old output named only `::async_trait` / `::serde_json` and stable `::genai_rs::function_calling::*` paths, so independent pinning was harmless before and isn't now. Documented in two places: the CHANGELOG entry, and the `genai-rs-macros` crate docs, since that's where someone reading about the macro would be when it bites them.

```
could not find __private in genai_rs
```

is the failure, and it's every `#[tool]` site at once, so at least it's loud rather than subtle. Only the macro-ahead-of-library skew breaks; the reverse is fine, as you noted.

**CHANGELOG.** Added — and you were right that there was no `[Unreleased]` section, which means my original edit had silently no-oped rather than landing somewhere wrong. I checked #446 and #448 for the same failure; both had it, both fixed.

**Two stale references.** Both taken, and both fall on the removable side of the line I drew:

- `docs/FUNCTION_CALLING.md:76` pointed at the README note this PR deletes, describing a requirement that no longer exists. Replaced with a pointer to the trybuild fixture, which is the thing that now keeps it true.
- `:113` and `:128` carried the import while only *defining* `#[tool]` functions. Same genuinely-unused category as the two I did remove — I'd matched on `.declaration()` presence per file rather than per snippet, so a file with any method-position call kept all of its imports. Removed.

**`Option<T>` branch.** Taken. This one mattered more than "more coverage": required and optional arguments go through different extraction branches, both were rewritten to route through the re-export, and only one of them was being compiled in the no-imports environment — so the fixture was half-testing the thing it exists to test. The fixture now takes an `Option<String>`, with a comment saying why it isn't decoration.

I'll leave the `json_value_to_tokens` numeric arms, as you suggested was fine — those vary by token kind rather than by import path, so they're not in this PR's risk surface.

```
cargo test --lib        858 passed
cargo test --doc        all doctests ran
cargo test --test ui_tests   4 fixtures, all ok
```

---

evansenter:
Applied in ed65f40, taking your first option rather than the reword.

You're right that the claim and its apparent counterexample sat in the same block. And the reason to prefer rewriting over rewording is the one you named: this is the only `rust,no_run` `#[tool]` snippet in the tree, so switching it to `get_weather_declaration()` turns the doc from an *assertion* about the import-free path into in-repo compile coverage of it. That's a strictly better guard than the comment I'd written, and it costs nothing.

The comment now also records that method position still works and what it costs, so the alternative style isn't erased — just no longer presented as a contradiction of the sentence above it.

This does not widen the ~15-snippet scope the PR body carves out: those remain `rust,ignore` and keep their imports because they make real method-position calls.

```
cargo test --doc   all doctests ran
```

Also noting your independent checks — particularly that `json!` is `$crate`-hygienic so routing it through the re-export is safe, and that pinning `async_trait` through the library is what kills the `E0195` rather than only the three `E0433`s. That second one I had observed in the before-state output but not explained; that's the mechanism.

---

evansenter:
Round 6 addressed in 9bdc19e (scope) and e15a8e7 (the two doc items).

**The three stray scripts — my error, removed.** You read it correctly as a stray cherry-pick. What actually happened: I was building the #428/#431 branch, switched here to handle round-5 feedback, and `git add -A` swept up the untracked files, since `git stash` had left them alone. They're gone from this PR and will land in the one that wires them up.

Your framing is the right one and I want to underline it: they'd have merged *without the review their own PR would have given them*, and the proof is in your other two comments — one of those scripts has a real bug and the other is inert.

**`export TITLE` after the query.** Confirmed, and this is a genuine defect, not a style point. `TITLE` is a plain shell variable at that point, so jq resolves `$ENV.TITLE` to `null`, `select(.title == null)` matches nothing, `EXISTING` is always empty, and the else branch always runs — meaning N consecutive failures file N duplicate issues, and the whole reopen/edit/"Still failing" path is dead code. The `--state all` comment describes exactly the outcome the bug produces, which is a nice tell. Your contrast with `resolve_scheduled_failure.sh` getting the ordering right is what makes it unambiguous. I'll fix it, plus the `mktemp` point, on the #428/#431 branch — and that branch also carries the `bash -n` gate from #438, which would not have caught this one (it's valid syntax that means the wrong thing) but is worth noting as a limit.

**`setup-dev.sh` referencing a nonexistent example file.** Same root cause; it only makes sense after the `.cargo/config.toml` → `.cargo/config.toml.example` rename, which is the substance of #428.

The two actual round-5 items — the "from other modules" qualifier that my previous commit message wrongly claimed to have fixed, and `serde_json` missing from the manual-impl note — are both in e15a8e7. I added an `assert` on those replacements this time so a non-matching edit fails loudly rather than passing silently, which is what let the first one slip.

---

PR #448, opened 2026-08-16
Model the generic `tool_call` step, and stop the MCP counters reading zero

### LINKED ISSUE #433: `Step::McpServerToolCall`/`McpServerToolResult` appear unreachable — the API emits generic `tool_call` steps

evansenter:
Closes #433.

## Re-probed before modeling

I ran the MCP call again live rather than working from the issue's transcript — 2026-08-16, `gemini-3.7-flash`, a real server at `mcp.deepwiki.com`, successful call:

```
status: completed
  tool_call     ['id', 'signature', 'type']
  model_output  ['content', 'type']
  thought       ['signature', 'type']
  model_output  ['content', 'type']
tool_use_tokens: 1962
```

Confirms the issue exactly: no `mcp_server_tool_call` step, and no server name, tool name, or arguments anywhere. (My first probe attempt used `server_label`/`server_url` and got `Unknown parameter` — the crate sends `name`/`url`, which is what the second probe used.)

## What landed — options 1 + 3, with 2 as the note

**1. `Step::ToolCall { id, signature }` is modeled.** These were falling into `Step::Unknown`, which is Evergreen working correctly but means the crate reported *no tool calls* on a successful MCP interaction.

**3. `step_summary().tool_call_count`.** This is the misleading part the issue identifies, and I agree it's worse than absent: `mcp_server_tool_call_count` returning `0` reads as "MCP didn't run" rather than "we can't tell you". Both fields now carry docs saying which one to check and why the other is zero.

**2. The MCP variants stay, documented as spec-only.** Same status as `Tool::Retrieval`. Deliberately *not* removed (option 4): unlike `cached_content` under D-005, nothing rejects these — they simply haven't been seen — so the endpoint may begin emitting them, and the sweep in #438 is what would tell us if the spec drops them.

## Tests

Two unit tests pin the observed payload:

- `tool_call_step_deserializes_the_observed_mcp_shape` — the exact three-key wire object, asserting it lands on the variant *and* explicitly that `is_unknown()` is false, since falling through to `Unknown` is the specific regression, plus a byte-for-byte roundtrip.
- `tool_call_step_allows_a_missing_signature` — the API sends it, the spec doesn't require it.

```
cargo test --lib     860 passed
cargo test --doc     all doctests ran
```

## What this does not fix

Which MCP server or tool was invoked is still not recoverable from the response. That's the API's shape, not the crate's — the call happened and `usage.total_tool_use_tokens` proves it, but the identifying detail isn't sent. `docs/ENUM_WIRE_FORMATS.md` now records that with the probe output rather than leaving a reader to infer it from a type that looks like it should carry the answer.

---


evansenter:
All four applied in be7fdb0.

**[Important] `Step::signature()` — confirmed, and this was a real bug I introduced.** The or-pattern didn't list `ToolCall`, so it fell through to `_ => None`. Your failure scenario is the right one: the live probe shows `signature` present on the observed `tool_call` step, so a caller collecting signatures for stateless replay would have silently dropped one the server actually sent — on the common case, not an edge. Fixed with the one arm, and pinned in the test that already covers the observed shape:

```rust
assert_eq!(
    step.signature(),
    Some("opaque"),
    "signature() must report it — the API sends one on every observed \
     tool_call step, and a caller collecting signatures for stateless \
     replay would silently drop it"
);
```

The `_ => None` catch-all is what let this through; adding a variant to that enum doesn't fail the build the way an exhaustive match would.

**`arb_step()` missing the variant.** Taken. Your point that `prop_oneof!` isn't exhaustiveness-checked is the reason it's worth an explicit note — I added the arm next to `file_search_call`, which has the identical shape, with a comment saying why nothing would have flagged its absence.

**StepDelta table row.** You're right that it implied a `StepDelta::ToolCall` that doesn't exist. Reworded to say outright that `tool_call` is a step type only, arriving whole on `step.start`, and what would happen to a `tool_call` delta if one ever appeared (`StepDelta::Unknown`, skipped with a warn). Your reading of the accumulator is correct — `step.start` carries the full `Step`, so the streamed array comes out right either way.

**CHANGELOG.** Added, including the `StepSummary` point — it isn't `#[non_exhaustive]`, so adding `tool_call_count` is source-breaking for an exhaustive struct literal or a destructure without a rest pattern. That's the kind of thing to read in release notes rather than discover at upgrade time.

On marking `StepSummary` itself `#[non_exhaustive]`: agreed it's defensible and agreed it's out of scope here. #446 is doing exactly that sweep for response types — `StepSummary` isn't in it because it's computed rather than deserialized, so my scanner doesn't see it. I'd rather it go in there than as a second unrelated break in this PR; noted on that PR.

```
cargo test --lib                        860 passed
cargo test --test proptest_roundtrip_tests   24 passed
cargo test --doc                        all doctests ran
```

---

PR #449, opened 2026-08-16
Make mold opt-in, and stop scheduled workflows failing silently

### LINKED ISSUE #428: Checked-in `.cargo/config.toml` hard-requires mold; a clone without it fails with a cryptic `cannot find 'ld'`

### LINKED ISSUE #431: Scheduled workflows fail silently: the flakiness report failed 22 days straight, then stopped running for ~10 weeks, unnoticed
Found while fixing #425, which asked whether the Apr–Aug gap in flakiness reports was a disabled schedule or a silently failing job. The run history says **both**, and neither produced any signal.

## The record

`actions_list` on `ci-flakiness-report.yml`, most recent 30 runs:

| Period | Runs | Conclusion |
|---|---|---|
| 2026-08-09 → 2026-08-16 | 8 | success |
| 2026-05-30 → 2026-08-08 | **0** | *(never ran)* |
| 2026-05-08 → 2026-05-29 | 22 | **failure** |

So the job failed 22 consecutive days in May, then stopped firing altogether for roughly ten weeks, then silently resumed on 2026-08-09. Nobody noticed any of the three transitions. The ~10-week silence is consistent with GitHub auto-disabling scheduled workflows after 60 days of repository inactivity, though I have not confirmed that specific cause.

## Why this matters beyond one workflow

The flakiness report is the repo's instrument for CI health. It being broken is exactly the condition it exists to surface, and it failed closed and quiet — the same class of problem as #421, where a doc asserted completeness it could not back. A monitoring job that can fail without telling anyone is worth less than its green runs suggest.

This generalizes: `audit.yml` (weekly) has the same exposure. Any scheduled workflow here can fail 22 times in a row without a human learning of it.

## What signal exists today

GitHub emails the workflow file's last committer on scheduled-run failure, which is easy to miss or filter and does not survive a change of maintainer. Nothing else fires. The 22 failures presumably generated 22 such emails and produced no action, which is itself evidence about that channel's effectiveness.

## Options

1. **Escalate to an issue on repeated failure.** A small step that, on failure, opens or updates a `ci-health` issue — the same mechanism the report already uses for its own output, pointed at its own health. Self-contained, no new secrets.
2. **Add a liveness assertion.** Have the report note when its own last successful run was older than ~48h, so a resumed-after-gap run says so in the body rather than quietly presenting fresh numbers as continuous.
3. **Notification integration** (Slack/email-to-list). Most reliable, but needs a secret and someone to own the channel.
4. **Accept and document.** Note in the workflow that failures are silent, so the next person reading a gap in reports knows to check run history first.

Leaning 1 plus 2 — 1 catches the failing-run case, 2 catches the not-running-at-all case, and neither has an owner-dependency. Option 3 is better if there is already a channel to hang it on.

## Also worth checking

- Whether the May failures had a single cause worth fixing, or were downstream of something since resolved. They stopped without an obvious fix commit, which suggests an environmental cause.
- Whether `audit.yml` has comparable gaps in its run history.

evansenter:
Closes #428, closes #431. Two CI/dev-environment issues, both "this fails without telling you".

## #428 — mold was a hard requirement

`.cargo/config.toml` was checked in with an unconditional `-fuse-ld=mold`, so a clone on a machine without mold failed every build before compiling anything:

```
error: linking with `cc` failed: exit status: 1
  = note: collect2: fatal error: cannot find 'ld'
```

`ld`, `lld` and `gold` are all present in that state — the missing binary is mold, which the message never names.

I have unusually direct evidence here: I hit this in a fresh container and carried `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS="-C link-arg=-fuse-ld=lld"` on every single cargo invocation for an entire working session. That is the cost the issue describes, paid in full.

**Option 1 from the issue.** The config ships as `.cargo/config.toml.example`, `.cargo/config.toml` is gitignored, and `./scripts/setup-dev.sh` enables it — after checking mold is actually present, printing per-distro install instructions and exiting 0 when it is not, since building without mold is fine and merely slower.

**CI keeps the speed-up.** The 13 existing "Install mold linker" steps across `rust.yml` and `release.yml` now copy the example into place alongside the install, so build times there are unchanged.

**Verified:**

```
$ cargo build --lib          # no override, container has no mold
    Finished `dev` profile [optimized + debuginfo] target(s) in 2m 14s
```

That exact combination was the reported failure.

## #431 — scheduled workflows fail quietly

The record from the issue: 22 consecutive failures in May, then ~10 weeks of not firing at all, then a silent resume. Three transitions, no signal anyone acted on. The only built-in mechanism is an email to whoever last committed the workflow file — the 22 failures presumably sent 22 of those.

**Options 1 and 2, minus the liveness half.** `CI Flakiness Report` and `Security Audit` (which the issue correctly flags as having the same exposure) now escalate a failing scheduled run into a rolling `ci-health` issue — the report's own mechanism pointed at the report's own health — comment on each subsequent failure so a repeat isn't swallowed by a silent body edit, and close it on recovery. Gated on `github.event_name == 'schedule'`, since a push/PR failure is already a red check.

### On the escalation scripts

These were briefly and accidentally included in #447 (a stray `git add -A`). That turned out to be useful: the review there found a real bug before it ever ran — `export TITLE` sat *after* the `gh issue list` whose `--jq` reads `$ENV.TITLE`, so it resolved to `null`, `EXISTING` was always empty, every failure would have filed a duplicate, and the entire update path was dead code.

Fixed, with a comment explaining the ordering, and now exercised against a stubbed `gh` covering both branches:

```
# existing issue present
GH-CALL: issue edit 77 --body-file /tmp/tmp.9wWcmoFFy9
GH-CALL: issue comment 77 --body Still failing: https://example/run/1
Updated issue #77

# no match
GH-CALL: issue create --title Scheduled workflow failing: CI Flakiness Report ...
```

Worth noting that `bash -n` — the gate added in #438 — would *not* have caught this. It is valid syntax that means the wrong thing.

Also switched the body file to `mktemp` rather than writing into the job's working directory, which is a checkout.

## Not done

The liveness assertion (option 2's "note when the last successful run was older than ~48h"). It needs run-history queries threaded into the report body, and it addresses the *not-running-at-all* case rather than the failing case. Worth its own change; this covers the half that has a clear implementation.

I also did not investigate whether the May failures had a single root cause — the issue lists that under "also worth checking", and the runs are past artifact retention.

---


evansenter:
**CI note:** `Integration Tests (multimodal)` is red on infrastructure, not on this diff — and I checked carefully rather than assuming, because this PR touches CI config and mold, so a linker regression is exactly what it *could* have been.

It isn't. The job died inside `dtolnay/rust-toolchain`, before any cargo command ran:

```
rustup toolchain install stable --profile minimal --no-self-update
info: syncing channel updates for stable-x86_64-unknown-linux-gnu
error: could not download file from
  'https://static.rust-lang.org/dist/channel-rust-stable.toml.sha256'
Caused by:
    1: error sending request for url (https://static.rust-lang.org/...)
    2: client error (Connect)
    3: Connection reset by peer (os error 104)
##[error]Process completed with exit code 1
```

Total job duration: 3 seconds. There was no toolchain, so `cargo` never ran, so `.cargo/config.toml` was never read and the mold change could not have participated.

Notably this is the second transient reset on this repo's runners in the last half hour — #446 hit the identical failure downloading `cargo-nextest` from GitHub releases. Different host, same signature, both in `westus2`.

I tried to re-run the failed job and got `403 Resource not accessible by integration`, so it needs a maintainer click or the next push. Flagging rather than leaving it silently red.

---

evansenter:
All five round-1 items applied in 416ba6e, including the blocker.

**[Important] The lookup would have degraded into the flood it prevents.** Confirmed and fixed. Your reasoning chain is the part worth restating, because each link is individually easy to miss: the flakiness report files one `ci-health` issue *per day*; `gh issue list` is newest-first; the rolling issue is created once and thereafter only *edited*; edits don't bump `created_at`. So it sinks steadily and silently falls off a 100-item page in about three months — at which point every failure files a duplicate and the close path stops closing anything. That is the same failure the `$ENV.TITLE` fix was made to prevent, arriving by a different route.

I took your second option: a dedicated `ci-health-escalation` label applied only by these two scripts, which keeps the searched population at one per watched workflow rather than depending on a limit at all. `ci-health` is applied as well so the issues still show in the existing view. Both lookups changed identically — your warning about drift is exactly right, since the failure mode there is updates landing on one issue while closes target another.

Worth noting this does not lean on #425 landing. If the daily cadence persists forever, the escalation lookup is unaffected.

**Composite action.** Taken, and your framing moved it from "pre-existing duplication" to "this PR made the copied block longer, and the `cp` is the half that goes stale on a rename". `.github/actions/setup-mold` now holds all three lines plus the timeout rationale — which was previously stated once as prose and eleven more times as "same stall ceiling as above".

The conversion asserted on the replacement count and caught its own shortfall: 12 of 13 on the first pass, because the cross-platform step has a different name and an `if:`. That one keeps its `if: runner.os == 'Linux'` on the `uses:` line.

**Host scope.** Taken. The message claimed the linker was enabled on any platform while the config only sets `[target.x86_64-unknown-linux-gnu]` — so an aarch64 user following the hints gets congratulated and sees nothing change. Now reads the host from `rustc -vV` and says so:

```
Enabled the mold linker via .cargo/config.toml (git-ignored).

Note: your host is aarch64-unknown-linux-gnu, and the config only covers
x86_64-unknown-linux-gnu — so builds will not actually use mold.
```

I also dropped the macOS hint on your `sold`-is-discontinued point; it was doubly wrong, since the config wouldn't have covered macOS regardless.

**Streak length in the issue body.** Not taken, and I want to be straight about why rather than let it look overlooked. You're right that the body edit is near-inert as written, and right that "failing since X, N consecutive runs" is the signal that would have made the original incident legible — `It failed again` ×22 is roughly what the email channel already gave. But deriving N from the comment count is fragile the moment a human comments on the issue, and `createdAt` gives first-seen without the count. I'd rather do it properly alongside the liveness assertion, which needs run-history queries anyway and is already deferred in the PR body. Filing it there rather than half-doing it here.

Also: thank you for verifying the `$ENV.TITLE` fix against real `gh` (`--jq '$ENV.HOME'` → `/home/runner`) rather than accepting my stub. That's the check I couldn't run.

All four `setup-dev.sh` paths exercised: no mold, mold on a matching host, mold on a stubbed aarch64 host, and a second run for idempotence.

---

evansenter:
**I broke CI and fixed it — recording both, since the previous comment claimed infrastructure and that was only half true.**

416ba6e took out ten jobs at once with:

```
##[error]Failed to load .../.github/actions/setup-mold/action.yml
   at GitHub.Runner.Worker.ActionManifestManagerLegacy.ConvertRuns(...)
```

Composite-action steps do not support `timeout-minutes:`, and an invalid manifest fails the workflow *before any step runs* — which is why it hit Check, Clippy, Documentation, MSRV, Coverage, Build Metrics and all three Test Suite jobs simultaneously rather than just the mold step.

Fixed in 0dff84d by bounding with the `timeout` command instead, so the stall protection survives — that bound is load-bearing, since a PR run once sat 98 minutes on an unresponsive apt mirror.

The lesson I'd have wanted earlier: I validated that the YAML *parsed*, which it did, and treated that as validation. The manifest was well-formed YAML and still rejected. The check now asserts against the keys a composite step actually accepts.

**Confirmed fixed on 0dff84d**, and the evidence is unusually clean:

| Job | Uses the composite? | Result |
|---|---|---|
| Check, Clippy, Documentation, MSRV | yes | **success** |
| Format | **no** (only cargo job without a mold step) | failed |

Every job that loads the composite now passes; the only failure is the one job that doesn't touch it. And that one is the `static.rust-lang.org` reset again:

```
rustup toolchain install stable --component rustfmt ...
error: could not download file from 'https://static.rust-lang.org/dist/channel-rust-stable.toml'
    3: Connection reset by peer (os error 104)
```

`cargo fmt --all -- --check` is clean locally. That's now three transient connection resets across this repo's runners in the last hour, across three different hosts (`static.rust-lang.org` ×2, GitHub releases ×1).

Still `403 Resource not accessible by integration` on re-runs, so Format needs a maintainer click or the next push.

---

evansenter:
Round 3/5 applied in `3e88054`. The blocker and all four suggestions are addressed; two of them I checked against primary sources and one of those inverted the premise.

**Blocker — `command -v mold`.** Replaced with a compiler probe: link a trivial C file with `cc -fuse-ld=mold`. Verified it discriminates rather than assuming:

```
cc -fuse-ld=mold        exit=1   (mold absent — this box)
cc -fuse-ld=lld         exit=0   (control: a linker that is present)
cc -fuse-ld=<bogus>     exit=1   (stands in for old-gcc rejection)
```

Missing mold and a driver too old to drive it both come back as a failed link, so they fold into the existing not-enabled branch as you said they would. The message now names both causes and prints the compiler's own error. All five script paths re-run.

**`rustc`-missing exit.** Reproduced — exit 127 after printing "Enabled the mold linker", exactly as described. `|| true`.

**apt budget.** Took the 60/240 split. You are right that one shared timer quietly collapses the `|| true` on `update`, and that is the more valuable property to keep.

**Composite `RUNNER_OS` guard.** In, with the restriction named in `description:`, and `cross-platform` drops its `if:`. Verified the skip path on `macOS`, `Windows`, and unset.

**`audit.yml` framing — your premise does not hold at v1.4.1.** I fetched the action's source rather than reasoning from the README's phrasing. `src/main.ts:85`:

```ts
if (github.context.eventName == 'schedule') {
    await reporter.reportIssues(actionInput.token, advisories, warnings);
} else {
    await reporter.reportCheck(actionInput.token, advisories, warnings);
}
```

The `schedule` branch files one issue per advisory and never calls `setFailed`. So a genuine RUSTSEC finding **leaves the scheduled run green** — a red one really is the audit breaking, and the monitoring framing was already correct. I still took the third-argument mechanism, because the inverse problem is real: someone arriving at a security-shaped issue title needs to be told the advisories live in separate issues, not this one.

That also resolves your `issues: write` note, and sharpens it. Without the grant, `reportIssues` 403s, the rejection reaches the top-level `catch`, and `core.setFailed` turns a real advisory into an opaque red run — so the permission was not merely inert, it was actively converting findings into noise. Documented as the two-consumer grant it is.

**`failure()` boundary.** Named in both workflows: not cancelled jobs, not execution-limit-exceeded, and the die-before-checkout case the script header already carried. All three want the same `workflow_run` watcher, deferred with the liveness work.

**Harness.** Both of your points were right. The drift check could not print its own failure — non-matching `grep` under `pipefail` took `set -e` before the FAIL branch — and comparing titles only proved "this literal appears in both files". It now compares `ESCALATION_LABEL` values, which genuinely differ on drift. Added streak, streak-reset and context-override coverage, and mutation-tested every new check rather than trusting a green run:

```
CAUGHT  - label typo in the resolver
CAUGHT  - streak ignores the Recovered reset
CAUGHT  - streak drops the opening run from the count
CAUGHT  - context override ignored (default always used)
```

**Streak length**, which I skipped in round 2 and should not have: the body edit was rewriting text identical bar a URL the comment already carried. Now "Failing since 2026-05-01 — 4 consecutive scheduled runs", derived from the issue's own comments since there is nowhere else to keep state, with a recovery resetting the count.

---

evansenter:
Round 6 applied in `f48af1a`. Two of the three are fixed; the third I want to answer rather than silently accept.

**Streak reset coupling.** Correct, and the sharper framing is that it fails *quietly in the field it was added to make trustworthy*. Both prefixes are now named constants, the reporter passes them into `jq` via `--arg`, and the harness asserts the two files agree — plus a second check that the resolver's comment is actually built from its own constant, since a constant can otherwise agree with itself while the emitted text has moved on. Mutation-tested with the exact rewording you named:

```
CAUGHT  - resolver reworded (the exact case round 6 named)
CAUGHT  - resolver comment bypasses its own constant
```

**`|| true` discarding exit status.** Also correct, and worse than "harness-internal": the two assertions it weakened are precisely the duplicate-filing guards, i.e. the regression this file exists for. A `run_script` helper keeps the tolerance where you said it belongs — off the substitution — and fails on non-zero:

```
CAUGHT  - reporter dies before the create-or-update branch
            FAIL - report_scheduled_failure.sh: the script exited 3
```

Under the old form that same mutation produced `ok - reporting: files no duplicate when one already exists`.

**`cross-platform` legs unverified before merge.** Accurate, and I am not going to widen the `push`-to-`main` gate to cover it — running three OSes × two toolchains in release on every PR is a real cost to buy one boolean. What I can say is narrower than "verified" and more than "untested": I ran the guard's shell body directly against `RUNNER_OS` of `macOS`, `Windows`, and unset, and all three take the early return with exit 0 before anything platform-specific executes. The residual is whether a composite `shell: bash` step runs at all on those images, which is a documented capability of all three hosted runners rather than something this PR is inventing.

Your comparison to the round-2 `timeout-minutes:` incident is the right one to make, but it cuts the other way: that failed at *manifest parse*, before any step ran, so it would have taken out the Linux jobs on this PR too — and did. This guard is inside a `run:` block that Linux exercises 13 times over. The untested surface is the two legs that skip, and a skip that fails is a red job on `main`, not a silent wrong result.

Recording it as accepted post-merge risk rather than closing it as covered.

---

PR #450, opened 2026-08-16
Stop describing by hand what release.yml already automates

### LINKED ISSUE #432: CLAUDE.md's Release Steps 3-4 describe by hand what release.yml already automates on tag push
`CLAUDE.md`'s **Release Steps** tells the releaser to create the GitHub release and publish both crates manually, but `.github/workflows/release.yml` triggers on `push: tags: ['v*']` and does both itself. Following the doc as written means racing your own CI.

## The drift

| CLAUDE.md | release.yml |
|---|---|
| 0. Local nightly docs.rs check | `validate` runs the same nightly check (still worth doing locally — see below) |
| 1-2. Tag and push | the trigger |
| 3. `gh release create vX.Y.Z --notes "..."` (copy from CHANGELOG) | `github-release` job, `softprops/action-gh-release` |
| 4. `cargo publish -p genai-rs-macros`, wait 30s, `cargo publish -p genai-rs` | `publish` job — same order, with a 30-attempt indexing poll instead of a fixed 30s |

Steps 3 and 4 are also listed in the opposite order from what the workflow does: the workflow gates `github-release` on `needs: publish`, so crates.io comes first and the GitHub release last. A releaser following the doc top to bottom would try to create the release before publishing.

## What actually happens if you follow it

Not a silent double-publish — `cargo publish` on an already-published version fails with `crate version ... is already uploaded`. But it fails *confusingly*, on a release that already succeeded, and the natural reaction is to go looking for a broken publish that isn't broken. Same for `gh release create` against a tag the workflow already released.

Verified on the 0.10.0 release: the tag push ran all three jobs unattended and published both crates (`genai-rs-macros` 15:26:10 → `genai-rs` 15:28:58 → GitHub release 15:29:15). No manual step was needed or possible.

## Step 0 should stay

The local nightly docs.rs check is not redundant even though `validate` repeats it. Its own comment explains why: catching it locally means finding out *before the tag exists* rather than after — and the tag is the irreversible part. That rationale is worth keeping verbatim.

## A real gap the doc reveals

CLAUDE.md says release notes are copied from `CHANGELOG.md`. The workflow does not read CHANGELOG.md at all — it writes a raw `git log --pretty=format:"- %s"` list and sets `generate_release_notes: true`. The v0.10.0 release body is:

```
- Release 0.10.0 (#416)
- Make the model a constant, bump the default to gemini-3.7-flash (#415)

## What's Changed
* Make the model a constant, bump the default to gemini-3.7-flash by @evansenter in ...
* Release 0.10.0 by @evansenter in ...
```

Two overlapping auto-generated commit lists, and none of the curated CHANGELOG prose. So the doc describes the *better* outcome and the workflow implements a worse one. Worth deciding which way to reconcile rather than just deleting the CHANGELOG mention.

## Proposed work

- [ ] Rewrite steps 3-4 as "the tag push triggers `release.yml`; watch the run" — validate, publish, github-release, in that order
- [ ] Keep step 0 and its rationale unchanged
- [ ] Note that `publish` re-verifies the tag against `Cargo.toml` version, so a mismatched tag fails before anything is uploaded
- [ ] Decide the release-notes question: either extract the `[X.Y.Z]` section from CHANGELOG.md into the release body, or drop the "copy from CHANGELOG" claim. Leaning the former — the CHANGELOG is already written by then, and the current body duplicates the same commits twice
- [ ] Consider whether the duplicated commit list + `generate_release_notes: true` should be just one of the two

Surfaced while releasing 0.10.0 (#416). Related to #426 only tangentially — that issue moves *rationale* out of CLAUDE.md and explicitly leaves operational checklists like this one in place, so this is not covered there.

evansenter:
Closes #432.

## The drift

`CLAUDE.md`'s **Release Steps** told the releaser to create the GitHub release and publish both crates by hand. `release.yml` triggers on `push: tags: ['v*']` and does both itself — and in the opposite order, since `github-release` needs `publish`, so crates.io comes first and the GitHub release last.

Following the doc top to bottom meant racing your own CI. Not a silent double-publish: `cargo publish` refuses an already-uploaded version, and `gh release create` refuses an existing release. But both fail *confusingly*, on a release that already succeeded, and the natural reaction is to go hunting for a broken publish that isn't broken. Verified against the 0.10.0 run — the tag push ran all three jobs unattended and published both crates; no manual step was needed or possible.

## What changed

**Steps 3–4 are now one step: watch the run**, with a table of what each job does in the order they actually run, plus the note that `publish` re-verifies the tag against `Cargo.toml` before uploading anything — so a mismatched tag fails before the first crate is pushed.

**Step 0 is unchanged**, including its rationale. The local nightly docs.rs check is not redundant with `validate` repeating it: catching a break locally means finding out *before the tag exists*, and the tag is the irreversible part.

## The release-notes gap the doc revealed

`CLAUDE.md` said release notes are copied from `CHANGELOG.md`. The workflow never read `CHANGELOG.md` at all — it wrote a raw `git log --pretty=format:"- %s"` list *and* set `generate_release_notes: true`, so the v0.10.0 body was two overlapping auto-generated commit lists and none of the curated prose.

Reconciled in the direction of the doc, since the CHANGELOG is already written by then:

- A new `.github/scripts/extract_changelog_section.sh` pulls the `## [X.Y.Z]` section out of `CHANGELOG.md` and that becomes the release body.
- `generate_release_notes: true` **stays** — it appends the PR-linked "What's Changed" list and the full-changelog compare link, which the CHANGELOG doesn't carry. Only the hand-rolled duplicate of that list goes away.
- A missing section **warns and falls back** to the commit list rather than failing. By that point the tag exists and both crates are published, so refusing to cut the GitHub release over a formatting problem would leave the release half-done.

## Testing

`release.yml`'s scripts otherwise only ever run at tag time — the one moment they cannot be debugged cheaply. So they are exercised on every PR:

- `.github/scripts/tests/test_extract_changelog_section.sh`, a fixture harness covering prefix collisions (`0.1.0` must not be satisfied by `[0.10.0]`, and a partial `0.1` must match nothing), em-dash and dateless headings, `[Unreleased]`, the final section in a file, an empty section, an absent version, and a missing file.
- A real-`CHANGELOG.md` extraction for the current `Cargo.toml` version. Warns rather than fails — between a version bump landing and its CHANGELOG entry being written, and on any PR predating the bump, the absence is the normal state.

Both run in the existing **`shell-scripts`** job rather than a new one. An earlier revision of this PR added a dedicated `release-scripts` job; that was dropped once #443 landed `shell-scripts` on main, because it would have been a third checkout-plus-run-a-script job with the same "no toolchain, reports in seconds" rationale — the pattern #443 consolidated and #441 deleted a copy of.

The harness needs no wiring at all: `make test-scripts` globs `.github/scripts/tests/*.sh`, so it is picked up by name, and it now also gets `shellcheck -S warning` coverage the dedicated job never gave it. The real-CHANGELOG step carries `if: ${{ !cancelled() }}` so it stays independent of the harness step — Actions stops a job at the first failing step, so without it a red harness elsewhere would silently skip the one check that reads the shipped file.

Mutation-tested: dropping the empty-section guard, the blank-line trim, or the stop-at-next-heading each fail the harness. One mutation (whole-heading equality → substring) was behaviour-equivalent rather than caught — the brackets are what prevent prefix collisions, not the equality — so the code comment was corrected to say that instead of claiming otherwise.

https://claude.ai/code/session_015jqCMGodSnv2THiRwzB1RK

---


evansenter:
Round 1 addressed in 5c1d8a8. All four taken.

**1. `::error::` dropped from the script.** Correct, and the release-side version of the problem is the sharper one: an error annotation on a release that worked is exactly the confusion this PR exists to remove. Both messages are plain stderr now, and each caller annotates at the level it actually means — `release.yml` warns and falls back, `rust.yml` warns for the reason its own comment gives.

**2. `tac` removed.** Also correct, and the failure mode you describe is the bad part: under the script's own `set -euo pipefail` the missing `tac` surfaces as a wall of empty-actual mismatches, not "command not found", so the natural read is a broken extractor. Trimming now happens in the awk pass that was already there — blank lines held pending, flushed only when a non-blank follows, which drops the leading run and the trailing one together. Single pass, no `tac`, no `sed`.

**3. The trailing assertion was vacuous — you're right, and the mutation-testing claim in the PR body was wrong with it.** Command substitution strips trailing newlines itself, so that line passed no matter what the extractor emitted, and what actually caught the dropped trim was the `assert_eq` above. Replaced with a check against the file the callers redirect into:

```bash
"$EXTRACT" "$work/CHANGELOG.md" 0.10.0 >"$work/body.md"
trailing=$(od -c "$work/body.md" | tail -2 | head -1)
case "$trailing" in *'\n  \n'*) fail "trailing blank line reached the file" ;; esac
[ "$(tail -c1 "$work/body.md" | od -An -c | tr -d ' ')" = '\n' ] || fail "..."
```

Both halves are now mutation-visible. Re-ran four mutations against the awk rewrite — flushing pending blanks at EOF, dropping the leading-run guard, dropping the empty-section guard, never stopping at the next heading — all four **caught**, where the first two previously would not have been.

**4. `grep | head -1` → one awk pass.** Taken for the reason given: this step turns on `pipefail` and the byte-identical `release.yml` line does not, so the SIGPIPE hazard is real here even though it isn't reachable against one matching line in today's `Cargo.toml`. `awk -F'"' '/^version = /{print $2; exit}'` gives the `head -1` semantics for free.

Harness passes, both workflows parse, every `run:` block passes `bash -n`, and the extractor still produces the 65-line 0.10.0 body ending in exactly one newline.

---

evansenter:
Round 2 addressed in fdd20d7. All three taken, and the first one was a genuine miss on my part.

**1. The replacement assertion still couldn't fail — correct, and my mutation claim was wrong twice over.** The stripping happens *inside* the extractor: `section=$(awk ...)` eats every trailing newline before the final `printf` ever runs, so moving the observation point to a file moves it past the stripping rather than before it. Which also means the "flush pending blanks at EOF" mutation I reported as caught was invisible; what caught the earlier, larger mutation was the leading-trim check, exactly as you diagnosed.

Fixed the way you suggested — trailing spaces survive command substitution where trailing newlines do not, so the blank line before `## [0.2.0]` in the fixture now carries two spaces, and a dropped trailing trim surfaces as a whitespace-only last line:

```bash
last_line=${out##*$'\n'}
case "$last_line" in
    *[[:space:]]*) ;;
    *) fail "trailing blank line was not trimmed; last line is [$last_line]" ;;
esac
```

Re-ran precisely the mutation that was invisible before (flush `pending` unconditionally instead of holding it): **harness exits 1**, and 0 with it reverted. So it's earning its keep now.

**2. `od -c | tail -2 | head -1` → `tail -c2 | od -An -c`.** Taken, and you're right that it's worse here than in `rust.yml`: the harness sets `set -euo pipefail` at the top, so an abort mid-run leaves no FAIL line and no summary — just a non-zero exit part-way through. The second weakness you noted is real too — one od line can't see a doubled newline straddling a boundary; two bytes can. `last_line` is now pure parameter expansion, so that check has no pipeline at all either.

**3. CLAUDE.md states the fallback.** Added right under the job table:

> If that section is missing — the `## [Unreleased]` → `## [X.Y.Z]` rename in the Version Bump Checklist above is what creates it — the job does not fail. It emits a `::warning::No [X.Y.Z] section in CHANGELOG.md` and falls back to a raw commit list, which is then worth replacing by hand. The fallback leaves no trace in the release itself, so a release body that looks like a bare commit list means that warning fired.

The "leaves no trace" half is the part that makes the doc pointer worth having — otherwise the only evidence is an annotation on a run nobody is looking at.

Harness passes and `bash -n` is clean.

---

evansenter:
Round 3 addressed in 257a7e5. All three taken — and the first one was right for the third round running, which is a fair thing to have kept raising.

**1. The doubled-newline arm was dead.** `od -An -c` pads each byte into a four-column field, `tr -s ' '` squeezes that padding to one space, so a pattern asking for two spaces between the escapes was unrepresentable. One space now. Mutation-tested by making the final `printf` emit two newlines: harness exits **1**, and 0 with it reverted.

Your secondary point landed too — the comment read as though the block tested the trim, which it cannot, because the extractor's own command substitution strips trailing newlines before that `printf` ever runs. It's a regression guard on the `printf`; the `$last_line` check above is what covers the trim. The comment now says which is which instead of implying one thing does both.

**2. The fenced level-two heading.** Took the fixture option. It's now pinned:

```
- before the fence

```markdown
```

…as the extracted body for a section containing a fenced block whose body opens with `## `. Plus a note in the script header. Agreed a fence state machine isn't worth it for a Keep a Changelog file — but you're right that truncation is the one failure of this script that looks *identical* to success (exit 0, no fallback, no warning), so turning an invisible property into a documented one is worth the ten lines. Same argument the empty-section case won.

**3. "byte-identical" was stale.** Rather than softening the word, I swept both `grep | head -1 | sed` sites in `release.yml` to the same awk one-liner — the cheap move you suggested. One rule with three readers that can't drift, and the load-bearing copy is `release.yml`'s since it gates publishing. Verified both produce `0.10.0` against this tree. The comment now says "the equivalent steps in release.yml" and notes the sweep.

Harness passes, both workflows parse, every `run:` block passes `bash -n`.

---

evansenter:
Round 4 addressed in 23bc06d. All three taken — and the first is the best catch of the four rounds.

**1. The invisible fixture.** The trailing-trim check depends on two literal spaces on one blank line of the heredoc, which is exactly the kind of thing a whitespace sweep or a save hook removes without anyone noticing — and the failure mode is silent reversion to the vacuous form rounds 2 and 3 were about. Pinned:

```bash
grep -q '^  $' "$work/CHANGELOG.md" ||
    fail "fixture lost its whitespace-only blank line; the trailing-trim check below is now vacuous"
```

Mutation-tested by stripping the spaces: the harness now fails with that message rather than passing. Given the last two rounds were both "this assertion stopped testing anything", making that condition itself an assertion is the right shape.

**2. `--pretty=format:` → `tformat:`, both sites.** Correct, and the framing is what makes it worth fixing: the two branches of the same `if` disagreed about the file they both write, and the disagreeing one is the branch nobody exercises before a real tag. Added a comment saying so, since `format` vs `tformat` is otherwise a one-character difference nobody would look twice at.

**3. CLAUDE.md wording.** Also right, and it sharpens the diagnostic rather than just adding detail — since the generated list is appended on the success path too, "looks like a bare commit list" was a signal a releaser couldn't actually use. Now: the body is the CHANGELOG section *followed by* the auto-generated PR list and compare link, and the fallback signal is the **absence of the CHANGELOG prose at the top**, not the presence of a list at the bottom.

Harness passes, both workflows parse, every `run:` block passes `bash -n`.

---

evansenter:
Round 5 addressed in 17818eb. Two taken in full; the third split.

**1. actionlint — ran it rather than wiring it in, and that turned out to matter.** Downloaded actionlint 1.7.7 + shellcheck 0.10.0 and ran them against this branch. **28 findings** across three workflows:

| Workflow | Findings |
|---|---|
| `rust.yml` | 16 |
| `ci-flakiness-report.yml` | 11 |
| `release.yml` | 1 |

So turning the gate on here would have reddened CI on 26 issues in files this PR doesn't own — and 11 of those are in `ci-flakiness-report.yml`, which #441 is concurrently rewriting, so they'd need fixing on two branches. Filed as **#453** with the counts and a note to sequence it after #441 and #443.

**But the one finding in this PR's own file was real and I've fixed it here:** `softprops/action-gh-release@v1`'s runner is too old for GitHub Actions. That's in the one workflow that only ever executes after a tag exists — a hard failure waiting for the least recoverable moment. Bumped to v2. That alone justified running the tool, and it's the argument for #453.

**2. `exit 2` no longer collapsed.** You're right that it was conspicuous next to a header comment explaining at length why the script declines to pick a severity — the one distinction it *did* make was being flattened by both callers. `release.yml` is the branch that ships a release body, so that's the one that separates them now: 0 uses the CHANGELOG, 1 warns and falls back, anything else is a hard `::error::` naming the exit code. Simulated all three paths (0 → step 0, 1 → step 0 with warning, 2 → step 1).

**3. The recovery advice.** This is the one where you found an actual trap: `publish` failing *between* the two crates cannot be recovered by re-tagging at all, because `genai-rs-macros` is already on crates.io and that version can never be re-uploaded — the retry fails on its first `cargo publish` and `genai-rs` never ships. Someone following "delete the tag, fix the version, tag again" would burn a cycle discovering that. Replaced the single paragraph with a table over all four failure points, including the `validate`-may-flake note and the "re-run first, re-tag only for a real defect" ordering.

Workflow parses, every `run:` block passes `bash -n`, harness passes.

---

evansenter:
Round 6 addressed in fcdb585. Two taken; the third is already landing elsewhere.

**1. The missing-file case shared exit 1 with the missing-section case — correct, and it defeated the dispatch I'd just added.** `release.yml` would have annotated a vanished `CHANGELOG.md` as "no [X.Y.Z] section in CHANGELOG.md", naming a section inside a file that doesn't exist, and shipped a commit-list body over it. Took the `exit 2` option: 1 is now reserved strictly for "the file is fine, the section is not there", which is the only case a caller should fall back on; a missing file is the same class of caller/environment problem as bad arguments. The header documents all three codes instead of only the recoverable one, and the harness pins all three explicitly rather than asserting non-zero, so the classification can't collapse again.

**I also gave `rust.yml`'s caller the same dispatch**, reversing the scope call you flagged as "considered" in the round-6 summary. Your reasoning here changed my mind: that caller had the identical misdiagnosis, and "unreachable in a checkout" is the same argument that was true of the release-side case being fixed. Consistency was cheaper than the distinction.

**2. The `## [Unreleased]` claim.** Right — the file on main has no such heading, so a releaser following that line finds nothing to rename and can reasonably conclude the step is done, which is precisely the state that triggers the fallback the paragraph is explaining. Reworded to cover both shapes: rename `## [Unreleased]` if present, otherwise add a `## [X.Y.Z] - YYYY-MM-DD` heading above the previous release. Left the Version Bump Checklist row alone since it's a compact table cell, and this paragraph is where the consequence is spelled out.

**3. `make test-scripts` + the CLAUDE.md row — already done in #443**, which adds exactly that target, globbed over `.github/scripts/tests/*.sh` and wired into `check:`, plus the command-table row. This PR's harness lands in that glob with no extra edit. Adding a second copy here would just conflict at merge, so I've left it to #443.

Harness passes (all three exit codes pinned), both workflows parse, every `run:` block passes `bash -n`.

---

evansenter:
Round 7 addressed in 388fd37. All three taken, and the middle one is the most serious thing found in this PR.

**1. The empty `VERSION` read in `rust.yml`.** Right, and it's the same shape as the round-4 fixture guard — make the precondition an assertion so the check can't quietly stop testing anything. `set -u` doesn't catch it because the variable is set and merely empty. Guarded with an `::error::` naming the cause.

**2. The same read in `release.yml` is worse, and you're right that the sweep didn't introduce it.** The value becomes a **grep pattern**, and an empty pattern matches every line — so the poll reports `Crate indexed successfully after 10 seconds` on its first attempt without the crate being indexed at all, and `genai-rs` then publishes against a dependency crates.io can't resolve. That's the exact race the 30×10s poll exists to close, silently disabled. Confirmed the `grep -q ""` behaviour rather than assuming it. Guarded, and the pattern is `-qF --` now since a version is data, not a regex.

Your note about the tag check being "safe by luck rather than design" is worth having recorded, so the commit says the three sites fail in different directions: one vacuous, one silently-passing, one incidentally-safe.

**3. The exit-2 branch hard-failing after publish — taken, and this was my mistake from round 6.** The job is `needs: publish`, so both crates are immutable by the time it runs: exiting produces precisely the half-done release (versions published, tag pushed, no GitHub release) that the fallback three lines below exists to prevent. I fixed the *diagnosis* last round and changed the *outcome* without noticing.

Now it annotates `::error::` and falls through to the commit list — a red annotation on a completed release rather than a completed publish with no release, which is your suggested split. Restructured so each exit code annotates at its own severity (0 → CHANGELOG, 1 → `::warning::`, else → `::error::`) and only the body generation is shared; otherwise the rc=2 case would have picked up the "No [X.Y.Z] section" warning too, reintroducing the wrong diagnosis. Simulated all three: bodies and exit codes as intended, step green in all three cases.

Harness passes, both workflows parse, every `run:` block passes `bash -n`.

---

PR #451, opened 2026-08-16
Send `InteractionInput::Content` as a `user_input` step

### LINKED ISSUE #427: `InteractionInput::Content` silently can't carry video `processing` — API rejects it outside a `user_input` step
Discovered while implementing #419. The API accepts video `processing` only when the content sits inside a `user_input` step; the bare-content-array input form is rejected.

## Reproduction

Identical content, two input forms, verified 2026-08-16 (`gemini-3.7-flash`, Api-Revision `2026-05-20`):

```jsonc
// A — bare content array (what InteractionInput::Content sends)
{"input": [
  {"type": "text", "text": "Describe briefly."},
  {"type": "video", "uri": "https://...", "processing": "static"}
]}
// → 400 Unknown parameter 'processing' at 'input[1]'.

// B — wrapped in a user_input step (what InteractionInput::Steps sends)
{"input": [{"type": "user_input", "content": [
  {"type": "text", "text": "Describe briefly."},
  {"type": "video", "uri": "https://...", "processing": "static"}
]}]}
// → 200 completed
```

The spec's input union is `str | [Step] | [Content] | Content`, and both arms are otherwise valid — so this reads as an API-side inconsistency rather than an intentional constraint.

## Why it matters here

`Content::with_processing()` is on `Content`, which callers naturally pair with `InteractionInput::Content` — the shorter of the two forms and the one most examples use. That combination always 400s. The failure is at least loud rather than silent, but it points the caller at the wrong thing: the message names the field, not the input shape.

## Options

1. **Document only** (what #419 shipped). Zero risk, but the footgun stays.
2. **Auto-wrap on the builder.** When input is `Content(..)`, serialize it as a single `user_input` step. Arguably the better default regardless — the steps form is canonical in revision 2026-05-20, and `Turn` was already removed in favor of steps. Needs live verification that the step form is accepted everywhere the bare form currently is, across text/image/audio/document, before it can be a blanket change.
3. **Auto-wrap only when needed.** Promote to a step only if some content carries `processing`. Narrower blast radius, but a request's wire shape then depends on a field buried in one content block, which is harder to reason about than either consistent choice.
4. **Reject at build time.** Return a builder error when `Content(..)` input contains `processing`, trading a remote 400 for a local error.

Leaning 2, with 4 as the cheap interim if a full verification sweep is not worth it yet. Option 3 is the one to avoid — surprise-by-context.

## Also worth reporting upstream

If there is a channel for Interactions API feedback (`docs/INTERACTIONS_API_FEEDBACK.md` suggests there is), this belongs in it: the same content object is valid in one input arm and invalid in another.

evansenter:
Closes #427.

## The problem

The API accepts video `processing` only when the content sits inside a `user_input` step. The identical content in a bare array — what `InteractionInput::Content` emitted — is rejected:

```
Unknown parameter 'processing' at 'input[1]'.
```

That message names the field rather than the input shape, so it points the caller at the wrong thing. And since `Content::with_processing()` is a method on `Content`, it naturally pairs with `InteractionInput::Content` — the shorter form, and the one most examples use. That combination always 400'd.

## The fix

Option 2 from the issue: `Content(..)` serializes as a single `user_input` step wrapping the blocks.

The issue asked for live verification that the step form is accepted everywhere the bare form currently is, before this could be a blanket change. Probed 2026-08-16 against `gemini-3.7-flash` at revision `2026-05-20`, sending identical content both ways:

| Input | bare content array | `user_input` step |
|---|---|---|
| text | completed | completed |
| image (inline base64) | completed | completed |
| audio (inline base64) | completed | completed |
| document (inline base64 PDF) | completed | completed |
| video (URI) | completed | completed |
| video + `processing: static` | **400** | completed |
| follow-up turn via `previous_interaction_id` | completed | completed |

So the step form is accepted everywhere the bare form is, and in one place the bare form is not. It is also the canonical shape under this revision — the one `Turn` was removed in favour of.

## Choices worth flagging

**Unconditional, not "only when some block carries `processing`."** That was option 3 in the issue, and it's the one to avoid: a request whose wire shape depends on a field buried inside one content block is harder to reason about than either consistent choice. The empty vector is wrapped too, for the same reason.

**Round-trip changes.** `InteractionInput::Content(..)` now deserializes back as `InteractionInput::Steps(vec[Step::user_input(..)])`, because the two are wire-identical. This is not new in kind — the existing proptest already documented that `Content(vec[])` came back as `Steps(vec[])` — but the shape it pins has changed, so that test is updated and extended to cover the non-empty case.

**No clone on the hot path.** The wrap goes through a borrowing `UserInputRef` rather than building a `Step` per serialization; a unit test asserts it is byte-identical to a hand-built `Step::user_input`.

## Testing

`cargo test --workspace` green (862 unit tests), `fmt`, `clippy --all-targets --all-features -D warnings`, and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` all clean. Four new unit tests cover the wire shape, equivalence with a hand-built step, the empty case, and that `Text`/`Steps` are untouched.

## Still open from the issue

The upstream report — the same content object being valid in one input arm and invalid in another — is not filed by this PR. Worth doing via whatever channel `docs/INTERACTIONS_API_FEEDBACK.md` describes.

https://claude.ai/code/session_015jqCMGodSnv2THiRwzB1RK

---


evansenter:
Round 1 addressed in 88b318f. Four of the five taken as written; the fifth stopped applying once the second landed.

**1. `Content::with_processing()` doesn't exist — correct, and the comment was describing an unreachable pairing.** `processing` is #419's work and isn't in `src/` yet. Both the code comment and the CHANGELOG now say so: today this is alignment with the canonical form, and the value is that #419 can land without a second wire-shape decision. A reader grepping for `with_processing` will no longer come up empty and confused.

**2. The response-echo nuance — taken, and it's the right structural call.** The wrap now lives on `InteractionRequest::input` via `serialize_with` rather than on the type's `Serialize`. Wrapping is a decision about authoring a request; `InteractionResponse::input` relays server data, and rewriting that into a shape the server never sent is the one place this genuinely cuts against preserve-data-roundtrip. Added a test pinning that the bare path is *not* wrapped, so the scoping can't quietly regress.

**3. `&[Content]` instead of `&Vec<Content>`** — done, construction site and `serialize_entry` unchanged as you predicted.

**4. The `arb_interaction_input` note — this one dissolved.** Its rationale is that an empty JSON array deserializes as `Steps(vec[])`, which was made obsolete by the *type-level* wrap. Scoping the wrap to the request field in (2) makes `Content(vec[])` serialize to `[]` again, so the note and the `1..3` bound are accurate as they stand and I've left them alone. What I did instead: restored `empty_input_array_deserializes_as_steps` verbatim (it documents real behaviour again) and added a separate request-side proptest over `0..3` generated content blocks, asserting the wrap is unconditional and byte-identical to a hand-built `Step::user_input`. So the empty case gets generated coverage, on the path where it now matters.

**5. `docs/ENUM_WIRE_FORMATS.md`** — fair, and the CLAUDE.md convention does call for it. Added a Quick Reference row and a details section carrying the full probe table (both shapes × text/image/audio/document/video/stored-follow-up, live 2026-08-16), plus the request-vs-response scoping and a note that `processing` isn't modeled yet. The evidence now lives in the document that exists to hold it.

Also caught while doing (2): the rustdoc cross-reference to the new private serializer failed `-D warnings` as a public-item-links-to-private-item error, so the field doc states the reasoning inline and points at `ENUM_WIRE_FORMATS.md`.

`cargo test --workspace` green (864), `fmt`, `clippy --all-targets --all-features -D warnings`, and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` all clean.

---

evansenter:
Round 2 addressed in a7cc7d9. All three taken.

**1. `Trigger::interaction` — you're right, and my claim was absolute when it shouldn't have been.** It's an `Option<InteractionRequest>`, so the field-level wrap rides along and a trigger created outside the SDK with a bare `[Content]` input does get reshaped on re-serialize. Narrowed the wording where the rationale lives, and listed it as the **third** roundtrip asymmetry on `Trigger::interaction`, whose doc already enumerates two — that's the right home for it, since anyone reading about the other two is exactly who needs this one. No defect follows: both shapes are accepted, and the step form is the one you want on the `TriggerUpdateParams` send path.

**2. The empty-content case — probed it rather than reasoning about it**, since it's the shape whose wire form changed most:

```
bare []                -> 400 Missing input.
empty user_input step  -> 400 Request has empty input.
```

Both rejected, differing only in the message, so the trade you were worried about — a clear 400 becoming a response to an empty prompt — doesn't happen. Added as a row in the table with that note.

The other half of your comment stands on its own though: `with_content(vec[])` does reach the wire where the text path returns `InvalidInput` locally. That's a builder-validation gap rather than a wire one, so I've filed it as #452 rather than widening this PR — it's orthogonal to the wire shape and wants its own unit test next to `test_interaction_builder_build_missing_input`.

**3. The `Step` row.** Chose the "note it" option: the details section now records that the probe exercised exactly one step tag, `user_input`, so the neighbouring row staying "pending live verification" reads as deliberate rather than stale. Promoting the whole `Step` row would over-claim — the other tags are still unexercised.

`cargo test --workspace` green (864), `fmt` and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean.

---

evansenter:
Round 3 addressed in 4b00993. Both taken.

**1. The trigger asymmetry is now pinned by a test.** Right that it was prose-only while its counterpart had `test_bare_input_serialization_is_unwrapped` behind it — an asymmetric guard, which is the shape that goes stale silently. `trigger_interaction_reshapes_a_bare_content_input` deserializes a `Trigger` whose `interaction.input` is a bare content array, asserts it reads back as `Content` (the shape the server sent), then asserts the re-serialized `Trigger` carries a `user_input` step. That's exactly the paragraph, as an assertion, sitting next to the other trigger roundtrip tests.

**2. Quick Reference row now lists all four arms.** `input_from_value` does handle a bare object — a lone content object to `Content`, a lone step object to `Steps` — so the row was summarising three-quarters of the section beneath it.

`cargo test --workspace` green (865 now), `fmt` clean.

---

evansenter:
Round 4 addressed in ac39409 and 11e555f. Both taken.

**1. "Requests always send the step form" overstated it.** Correct — `serialize_request_input` rewrites only the `Content` arm, and `test_wrap_does_not_touch_text_or_steps_input` in this same PR asserts `Text` still goes out as a bare string, so the doc and a test were disagreeing. Both sites restated as what the code does: on requests the bare `[Content]` form is never sent, `Text` is unaffected, `Steps` was already the step form.

**2. The round-trip claim now has a direct test.** It was true only by composition — the serialize test plus `test_interaction_input_steps_array_roundtrip` — which is exactly the arrangement where a documented sentence can go stale without any test failing. `test_request_content_input_round_trips_as_a_user_input_step` serializes a `Content`-input request, deserializes the whole `InteractionRequest` back, and asserts the input is `Steps` holding one `UserInput` carrying the original content.

`cargo test --workspace` green (866), `fmt` clean.

---

evansenter:
Round 5 addressed in 16496a4 (which also picks up the two suggestions from the parallel round-5 review). Four items, all taken.

**1. The catch-all arm.** Taken, and the reasoning is right — `InteractionInput` is `#[non_exhaustive]` precisely because the input union grows, so defaulting future arms by omission is the thing that goes stale quietly. Now `Text(_) | Steps(_)`, which still compiles in-crate and will break the build *at this site* when an arm is added.

**2. `**BREAKING**` marker.** Added, with a half-sentence saying why it's marked despite no signature change: a caller who persists a built `InteractionRequest` and matches on `input` after reloading now takes a different branch, and the marker is what someone scans for when upgrading. Better to over-mark than to leave the classification to inference.

**3. The `Resolution` example.** Good catch — it's two hundred lines below the section this PR added, showing exactly the shape that section says is never sent. Nested under a `user_input` step, with a note that `resolution` still sits on the content block so the example doesn't lose its own point. Checked the rest of `docs/`: `CONVERSATION_PATTERNS.md`'s request example already shows the steps form, and the only other `"input"` occurrences are a string input and a curl snippet, so those two were the whole set.

**4. The builder exclusion.** You're right that the stated reason is now wrong — composing them would serialize as history steps followed by a `user_input` step, which is what the error tells the caller to build manually. Re-worded the comment to say it's a builder policy rather than a wire constraint, left the check alone, and filed **#454** for the actual API question (including the wrinkle that `with_content()` + `with_text()` already interact in a non-obvious way, so a third pairing is more surface than it first looks).

`cargo test --workspace` green (866), `fmt` and `clippy --all-targets --all-features -D warnings` clean.

---

evansenter:
Round 7 addressed in e137ce4.

Taken — and the asymmetry is what makes it worth fixing rather than a pure nit: the `Steps` bullet already carried a shape note while `Content` didn't, so the list read as though only one of them had anything to say about wire form. Since that list is what rustdoc renders first on the enum page, a reader who stops there came away with the pre-#427 model even though the variant doc below it was correct.

```
/// - `Content`: Array of content blocks for multimodal input — sent as a
///   single `user_input` step, not as a bare array (#427)
```

`cargo fmt --check` and `RUSTDOCFLAGS="-D warnings" cargo doc --all-features` clean.

---

